PR #9, opened 2015-08-27
Possibility to add multiple Licenses in missingFile

TobiNet:
Right now you can only add a single license to the properties File. Here you can add multiple, splitted with | to the File.

### DIFF
--- a/src/main/java/org/codehaus/mojo/license/api/ThirdPartyTool.java
+++ b/src/main/java/org/codehaus/mojo/license/api/ThirdPartyTool.java
@@ -147,7 +147,7 @@ public interface ThirdPartyTool
      * @param project     the project
      * @param licenseName the name of the license
      */
-    void addLicense( LicenseMap licenseMap, MavenProject project, String licenseName );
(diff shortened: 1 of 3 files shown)


src/main/java/org/codehaus/mojo/license/api/ThirdPartyTool.java
```diff
@@ -147,7 +147,7 @@ SortedProperties loadUnsafeMapping( LicenseMap licenseMap, SortedMap<String, Mav
      * @param project     the project
      * @param licenseName the name of the license
      */
-    void addLicense( LicenseMap licenseMap, MavenProject project, String licenseName );
+    void addLicense( LicenseMap licenseMap, MavenProject project, String... licenseName );
```
tchemit: Could you use **licenseNames** parameter please

TobiNet: Yes, thanks, forgot the interface!

tchemit:
@TobiNet Thanks for contribution  ;)

Issue #13, opened 2015-09-22
ID mismatch when using SNAPSHOT versions


philandrews-ixxus:
There is a problem when using SNAPSHOT dependencies which resolve to specific artifacts using maven-metadata.xml:
```xml
<snapshotVersion>
<extension>jar</extension>
<value>3.0.2-20150918.080710-5</value>
<updated>20150918080710</updated>
</snapshotVersion>
```
 which results in the dependency not being recognized when present in the missing file and removed, but then since no license is present in the POM the same dependency is added back to the missing file using the original ID which doesn't work.

The problem seems to be tied to an ID mismatch between the artifactCache map and the unsafeMappings map in loadUnsafeMapping in DefaultThirdPartyTool.

It looks like the resolved version comes in during loadProjectDependencies, but when the MavenProject is built by buildFromRepository, that object is created with the SNAPSHOT version but added to the cache with the resolved version.  So when it tries to map between the MavenProject in the unsafeDependencies set and the artifactCache, the lookup fails and the mapping is removed.  But then since the dependency remains in the unsafeDependencies set the mapping gets recreated using the ID that will never map.

kluckow:
Hey @philandrews-ixxus,
I am facing the same issue concerning those SNAPSHOTS with maven-metadata.xml's.

I have already tried some modifications concerning the MavenProject checks and tested it, but it did not work out and I have no idea if I am on the right way.

Do you have any idea for a solution to include SNAPSHOT+maven-metadata.xml handling?

kluckow:
I am pretty sure I got a solution for this problem.
Could anyone tell me how to contribute it the correct way? :)

kluckow:
Pull request sent...

tchemit:
Closed by PR #26

PR #25, opened 2016-02-02
ID mismatch when using SNAPSHOT versions #13

kluckow:
According to the detailed description of the problem...
That should convert any timestamp notation of snapshots into snapshot-notations.
So the plugin does not differentiate between those.


kluckow:
Can anyone explain what went wrong here? I dont really get it.

PR #26, opened 2016-02-02
ID mismatch when using SNAPSHOT versions #13

kluckow:
According to the detailed description of the problem...
That should convert any timestamp notation of snapshots into snapshot-notations.
So the plugin does not differentiate between those.


tchemit:
Hi @kluckow, PR was just merged.

I squash your commits, reformat code and use better code.

Thanks for your contribution.

PR #27, opened 2016-02-15
add option to add empty line after the header

maggu2810:
If you use this plugin to add the license header in front of the java
package line, there is no empty line inserted.
The package statement is in the next line after the closing header.
Checkstyle requires an empty line after a header / before the package
line.

Signed-off-by: Markus Rathgeb <maggu2810@gmail.com>


maggu2810:
I would like to change that, that it is only done for "in front of package line" and also if there is no empty line in front of already...

PR #28, opened 2016-02-18
Fix for String index out of range: -1.

ceagan:
I ran into a problem where the license-maven-plugin would fail when processing sql files producing an error _String index out of range: -1_. This error turned out to be caused by line 407 in _AbstractFileHeaderTransformer.java_ where the length of `s` was 2 because in the sql headers, the Comment Line Prefix is `--`. In Java files, the Comment Line Prefix is `*`, which is only of length 1 and does not cause the error. Swapping the contents of the if blocks results in the proper behavior.

### DIFF
--- a/src/main/java/org/codehaus/mojo/license/header/transformer/AbstractFileHeaderTransformer.java
+++ b/src/main/java/org/codehaus/mojo/license/header/transformer/AbstractFileHeaderTransformer.java
@@ -399,12 +399,12 @@ public abstract class AbstractFileHeaderTransformer
                 {
                     if ( line.length() <= s.length() )
                     {
-                        line = "";
+                        line = line.substring( s.length() );
                     }
                 }
                 else
                 {
-                    line = line.substring( s.length() );
+                    line = "";
                 }
             }
             buffer.append( line ).append( LINE_SEPARATOR );


stefanbirkner:
Could you provide a text for the problem?

ceagan:
I'm not clear on how the tests in the unit testing for this plugin are set up. There isn't a set of sample files which would be used to verify the functionality. If you can point me in the right direction, I can try to build a unit test for the functionality.

I have a very simple .sql file in my project which is what caused this problem. Note that the issue is with processing the header so if the file doesn't contain a header to begin with, the issue does not appear. You have to run the license-maven-plugin twice. One to generate the header into the sql file the first time and then again to cause it to fail while reading the header and unboxing it.

```sql
---
-- #%L
-- project-name
-- %%
-- Copyright (C) 2015 Company
-- %%
-- LicenseText
-- #L%
---

-- Create user
CREATE USER username WITH PASSWORD 'username';
-- Create database
CREATE DATABASE dbname OWNER username;
-- Grant user access to the database
GRANT ALL PRIVILEGES ON DATABASE dbname to username;
```

ceagan:
@stefanbirkner Is there something I can do to help get this change integrated?

stefanbirkner:
Thanks for the example. I started to write the test.

tchemit:
Hi @stefanbirkner what about the IT you began to write ?

tchemit:
@ceagan I can't reproduce your error. Can you attach a maven projet which prove the defect ? thanks.

ceagan:
I created an example with the issue here: https://github.com/ceagan/license-issue

tchemit:
Thanks for the patch :+1:

PR #29, opened 2016-03-02
add option to force an empty line after header

maggu2810:
If the header is in front of the package line ("addJavaLicenseAfterPackage=false") and you are using Checkstyle for your Java code, it complains that there should be an empty line in front of the package line.

Parameter: license.emptyLineAfterHeader

If this parameter is set to false (default) the behavior keeps the same as before.

If this parameter is set to true and there is a non-empty line after the header (the line is trimmed to check if empty) then an empty line is added between the header and the content.


maggu2810:
@tchemit Hi, what do you think about?

PR #32, opened 2016-05-03
Fixing license downloader to account for URL redirects, and checking the mime type for the correct file extension

rchargel:
Fixing license downloader to account for URL redirects, and checking the mime type for the correct file extension

### DIFF
--- a/src/main/java/org/codehaus/mojo/license/utils/LicenseDownloader.java
+++ b/src/main/java/org/codehaus/mojo/license/utils/LicenseDownloader.java
@@ -27,6 +27,7 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.net.HttpURLConnection;
 import java.net.URL;
 import java.net.URLConnection;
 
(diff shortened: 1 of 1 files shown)


tchemit:
Hi @rchargel , thanks for this PR. I will do some little modification on it, have a look at it if you want.

Could you explain me the big picture around the file extension and mime-types ? I don't get it.

rchargel:
I was having issues where a lot of license files were actually PHP pages. These PHP pages generated valid HTML, but the file extensions were still PHP. I wanted to have a way to replace the extension with the file extension that made the most sense for the returned mime type. 

For instance, a PHP page would be .html since it was valid HTML.
A page with no extension would use the mime-type to determine the most appropriate extension.

And finally if there wasn't a direct mime type --> extension link, it would just use the default extension.

Certainly, that was not a critical fix, but more of a nice to have for me.  The redirects from 302 and 301 responses were much more important.

tchemit:
Ok I will merge it like this but we may improve the extension system later :)

tchemit:
PR Merged, thanks for your contribution.

PR #33, opened 2016-05-03
Added the ability to organize licenses by the actual artifacts.

rchargel:
We needed a way to download all of the licenses from each dependency, and provide those licenses based on the dependency.

With the default implementation, we had 2+ dependencies with the same license name (specifically BSD) but the licenses were slightly different and we did not get both copies.

For our product, we are required to distribute all of the licenses for all of our dependencies. Even if they are the same (or similar) license.


tchemit:
PR merged, thanks for your contribution.

Issue #56, opened 2016-09-23
different line endings on windows and linux


heapifyman:
We have a problem with the license-plugin messing up line endings when adding the license header to source files in a mixed windows / unix environment using git for version control.

License information uses unix line endings and comes from a dependency artifact using  `<licenseResolver>classpath:`.

I also have configured `<canUpdateCopyright>true</canUpdateCopyright>` and `<canUpdateDescription>true</canUpdateDescription>`.

Git is configured on Linux machines with `core.autocrlf=input` on Windows with `core.autocrlf=true`, thus sources are committed with Unix line endings and converted to Windows line endings when cloning to a Windows machine - as recommended.

Now, when the license-plugin's update-file-header is run on a windows machine and the source file already has a license header, the license header will be replaced - in fact, only the line endings will be changed from windows style to unix style.
That in turn lets git think the file has been changed, which leads to the situation, where git and our IDE show every source file with a header as changed, although you haven't actually changed the code.
As you can imagine, that makes it quite hard to work with git because you can't really see what actual changes will be included in a commit.

Is there a way we can change this behaviour? Have we maybe misconfigured the plugin?
As a workaround we now have disabled the license-plugin on windows machines via maven profile but that's hardly an ideal solution.

Any hints would be appreciated.
Thanks in advance.

P.S.: I have used the command described at http://stackoverflow.com/questions/121392/how-to-determine-the-line-ending-of-a-file/2065154#2065154 to determine the line endings.

mpkorstanje:
Looks like the line endings are read directly from the license files. Additionally a few line endings \n are added when strings are concatenated.

heapifyman:
Wow, that was fast.
One question though, shouldn't the "\n" used in the different FileHeaderTransformers also be replaced?

mpkorstanje:
The buffered reader effectively splits the string on `(\r)|(\n)|(\r\n)` which covers old-mac, unix and windows respectively. The way line endings are added no longer matters.

Should be noted that this changes all the line endings in the file to be platform appropriate, not just those in the header as the plugin has done before. I don't see any problems with that though.

mpkorstanje:
Looks like I'm tripping up some tests though. :S

mpkorstanje:
Also worth noting that the groovy integration tests are now platform specific. Not so groovy.

mpkorstanje:
Fixed those too. 

@tchemit Everything should be good to go now.

heapifyman:
[...]

PR #57, opened 2016-09-23
 Print strings using platform specific line endings

mpkorstanje:
Resolves "issues different line endings on windows and linux #56" by  using a BufferedReader to take care of the possible line encodings and using a print writer to add the correct platform specific line endings.

 Renamed the method to printString to reflect the distinction between printing and writing as observed by java.

Apologies for any formatting mistakes. The formatting seems to be rather non-standard.


tchemit:
Thanks for the patch :+1:

PR #67, opened 2017-01-06
update to fix for #56

heapifyman:
use LINE_SEPARATOR instead of "\n"


tchemit:
thanks for the patch.

PR #79, opened 2017-03-08
Allow download-licenses mojo to write license summary without downloading licenses

eirikb:
Removing the mojo skip when offline is set to true.
Preventing downloadLicenses method from being called when offline is true,
this way the summary will be built, without any accompanying licenses.


tchemit:
thanks for the patch.

PR #82, opened 2017-04-03
Adds a filter to exclude transitive dependencies from excluded artifacts

dschroedter:


tchemit:
thanks for the patch.

PR #87, opened 2017-05-12
Fix PR-33 integration test

laurentgo:
License file for project ToPIA has been removed from website.
Fix the test by using a different URL.


tchemit:
Arch I did not see with PR, and I just fix it by my self, anyway thanks for the patch.

PR #89, opened 2017-05-16
Clarified doc for `fileTemplate` property in `add-third-party` mojo

rbarbey:
Clarified aspect of being both a file and a resource on classpath which might require an absolute path in case of running mojo within a sub-module of a multi-module build. Ran into this problem this morning and wanted to fix in the code but didn't want to mess too much with `DefaultThirdPartyTool` or the `FreemarkerHelper`. It would have required to use an instance of `MavenProject` itself which felt a little too heavy-weight.


tchemit:
thanks for the patch.

Issue #77, opened 2017-03-02
Override the identified during license-check


danielwegener:
As discussed in #59 it is sometimes necessary to override the license for a certain dependency. The fix created for #59 however does this only for the license-report. This overwrite should also be applied during license checks.

Example:

I get the following error (with `<failIfWarning>` is `true`):
```
[WARNING] There are 1 forbidden licenses used:
[WARNING] License CDDL + GPLv2 with classpath exception used by 1 dependencies:
 -javax.transaction API (javax.transaction:javax.transaction-api:1.2 - http://jta-spec.java.net)
```
Now I have created the file `my-proj/src/license/override-THIRD-PARTY.properties` with the following content:
```
# original: CDDL + GPLv2 with classpath exception
javax.transaction--javax.transaction-api--1.2=BSD-3-Clause
```
... while my project pom has an whitelist entry for `BSD-3-CLAUSE`:
```
<configuration>
 <includedLicenses>
   <includedLicense>BSD-3-Clause</includedLicense>
```

However, the build still fails with the same error as shown above.

solomax:
`aggregate-add-third-party` goal also would benefit from ability to override wrong/incomplete licenses

tchemit:
As said by @solomax , add-third-party and aggregate-add-third-party mojo use this mecanism, could you give us a little project that prove the defect @danielwegener , thanks.

diegosalvi:
[defect-77.tar.gz](https://github.com/mojohaus/license-maven-plugin/files/1559278/defect-77.tar.gz)

Mini project attached.

Run `mvn install` on it.

* It has a single dependency with license "Apache License, Version 2.0" (merged as AL v2 just to have small strings but it's the same without merges).
* It define a license override for its only dependency to dual "OVERRIDDEN" and "UNKNOWN"
* It accept only "OVERRIDDEN" licenses

Expected path should be:

"Apache License, Version 2.0" -> (merge) "AL v2" -> (override) OVERRIDDEN -> (check included) ok in whitelist.

But real path is:
"Apache License, Version 2.0" -> (merge) "AL v2" -> (check included) not in whitelist!!!

If failIfWarning is false the process continue and you'll find (OVERRIDDEN) (UNKNOWN) in the third party file as expected

ppalaga:
@danielwegener which mojos is this targeting?

It would be welcome to transform the attached project into an integration test under https://github.com/mojohaus/license-maven-plugin/tree/master/src/it

PR #90, opened 2017-05-22
Override the identified during license-check #77

diegosalvi:
I needed to resolve whitelist and blacklist using overridden licenses.

Just added method to encapsulate overrideLicenses call and invoked in both add-third-party and aggregate-add-third-party before license checkup


diegosalvi:
About tests note that the failing PR-33 fails on master too with the same error (the PR doens't introduce that error)

tchemit:
thanks for the patch.

PR #99, opened 2017-07-16
Add trimHeaderLine option

laurentgo:
Add trimHeaderLine option to update-file-header goal.
If the header contains an empty line, when boxed the output line
might contain trailing whitespace which might be rejected by
Checkstyle plugin for example.

The option will do a right trim before writing the line.


tchemit:
thanks for the patch

PR #101, opened 2017-07-18
Use test scope for junit dependency

ctrueden:


tchemit:
Thanks for the patch.

Issue #92, opened 2017-06-29
update-file-header changes file encoding


heapifyman:
We have a file in `UTF-8` encoding with special character German `ß` in javadoc.

We are using both Linux and Windows machines for development.

Everything works fine on Linux but when running `update-file-header` on a Windows machine, it changes the file encoding from `UTF-8` to Windows default, which screws up the `ß`, which in turn leads to a failure when creating the javadoc.

`license-maven-plugin` should respect the file's current encoding and not change it.

tchemit:
@heapifyman , Ok I just fix it, I will deploy a snapshot can you check the fix is ok form you ?

tchemit:
Snapshot 1.14-SNAPSHOT is ready to test :).

maggu2810:
I am not using a Windows machine myself but some co-worker.
There is a degree character (`°`) in a source file.
Without the change (fd20eb4294c00b325cd1d422d2c942c678e96870) the plugin changes the character and the compiler will break because of an invalid UTF-8 encoding.
With the change (fd20eb4294c00b325cd1d422d2c942c678e96870) the plugin changes the character but the compiler does not complain an invalid UTF-8 encoding.
So, the encoding seems to be correct, but the character is changed to an invalid one.

Perhaps someone with an Windows machine could test.

heapifyman:
@tchemit  Sorry for the late reply, I was on holidays.

Anyway, I just tested on a Windows machine and see the same behaviour as @maggu2810 :
with version 1.14-SNAPSHOT the build (and javadoc generation) succeeds but the character `ß` is still changed.

PR #102, opened 2017-07-26
keep encoding

maggu2810:
Fixes: https://github.com/mojohaus/license-maven-plugin/issues/92


maggu2810:
it does not work as expect, as it regenerates the license header

tchemit:
thanks for the patch.

Issue #103, opened 2017-07-28
Add *.kt (Kotlin) extension to JavaFileHeaderTransformer


comdiv:
For now i cannot update headers for kotlin files - only with custom rebuild

PR #104, opened 2017-07-28
Kt extension added for JavaFileHeaderTransformer for Kotlin (#103) Fixed

comdiv:


tchemit:
Thanks for the patch.

PR #108, opened 2017-10-10
Added licenseMergesFile user property

v1v:
I need to be able to define **licenseMerges**  as User property by command line to avoid touching the pom.xml file.

One optional properties was added to the 'license:add-third-party' and therefore 'license:aggregate-add-third-party'

**licenseMergesFile**
Load file supplying information for merging licenses in final file. If defined then it gets precedence to the licenseMerges property.
_User_ property is_: license.licenseMergesFile.

What do you think? Is this something others might get benefits?

Thanks


v1v:
Raised https://github.com/mojohaus/license-maven-plugin/issues/109 to ask why that particular jdk is failing

Issue #115, opened 2017-12-14
Cleaner third-party aggregate on multi license projects


diegosalvi:
Original grouped third-party file third-party-file-groupByLicense.ftl is annoying when multilicensed libs kicks in (each is listed multiple times).

A new third-party file with a cleaner listing could be helpful.

Like:

```
(license A, License B)
    libA
    libB
    ...

(license A, License C)
    ...

(license B)
    ...

```

ppalaga:
The fix seems to have been merged. Feel free to reopen if that's not the case.

PR #116, opened 2017-12-14
Multi license aggregate third-party for #115

diegosalvi:
Added a new "cleaner" aggregate file for resolution of #115


tchemit:
thanks for the patch :)

PR #118, opened 2018-01-11
Fix #117 Add possibility to include/exclude types

### LINKED ISSUE #117: Possibility to exclude/include certain artifact types
Our case are BOM dependencies which group dependencies together. These are there just for convenience, but are not important when it comes to the final assembly which includes all "relevant" licenses.
Currently it is not possible to filter for specific types, in our case to exclude BOMs, e.g.
```
         <execution>
            <id>exclude-bom</id>
            <goals>
              <goal>add-third-party</goal>
            </goals>
            <phase>package</phase>
            <configuration>
              <excludedTypes>pom</excludedTypes>
            </configuration>
          </execution>
```

I can provide a PR in the next days for that.

Matze2:
This PR allows to include/exclude artifact types in add-third-party and download-licenses.
My use case is to exclude "pom" dependencies that are not relevant for the final license compilation since they do not provide artifacts themselves.

### DIFF
--- a/src/it/add-third-party-excluded-included/pom.xml
+++ b/src/it/add-third-party-excluded-included/pom.xml
@@ -63,6 +63,12 @@
       <version>4.8.2</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.wildfly</groupId>
+      <artifactId>wildfly-ejb-client-bom</artifactId>
+      <type>pom</type>
+      <version>9.0.1.Final</version>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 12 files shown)


Matze2:
The travis configuration for "oraclejdk7" is wrong which makes the build fail.
The other platforms work correctly.

tchemit:
ITs are down, I will check that is the problem.

tchemit:
I found the problem (using junit 4.1 has a license which point to a 404 pb already solved AFAICR...

tchemit:
So I can cleary merge this one ;)

tchemit:
Thanks for the patch

Issue #109, opened 2017-10-10
Travis oraclejdk7 build seems to be broken (environmental issue)


v1v:
JDK: oraclejdk7 build seems to be corrupted

``` bash
The JAVA_HOME environment variable is not defined correctly
This environment variable is needed to run this program
NB: JAVA_HOME should point to a JDK not a JRE
```

Further details: https://travis-ci.org/mojohaus/license-maven-plugin/jobs/286028333#L455-L459

v1v:
https://github.com/travis-ci/travis-ci/issues/7964

PR #119, opened 2018-01-17
GH-109 oraclejdk7 is not supported anymore 

v1v:
#109


tchemit:
Thansk for the PR :) I am back on the project, after some long weeks of hard work in other places.
I will try to treat as much issues the week and try to produce a new version next week.

Issue #121, opened 2018-01-24
Broken TestIT


v1v:
As far as I see the current PRs are broken since http://www.opensource.org/licenses/cpl1.0.txt is not reachable and two ITs are failing over and over:
- ISSUE-40/postbuild.groovy
- download-licenses-basic/postbuild.groovy

As you can see below:

``` bash
[INFO] --- license-maven-plugin:1.15-SNAPSHOT:download-licenses (default-cli) @ it-test ---
[WARNING] Unable to retrieve license for dependency: junit:junit
[WARNING] http://www.opensource.org/licenses/cpl1.0.txt
[WARNING] Server returned HTTP response code: 403 for URL: https://opensource.org/licenses/cpl1.0.txt
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 9.189 s
[INFO] Finished at: 2018-01-24T14:52:51Z
[INFO] Final Memory: 16M/217M
[INFO] ------------------------------------------------------------------------
Running post-build script: /Users/victor/work/v1v/license-maven-plugin/target/it/ISSUE-40/postbuild.groovy
/Users/victor/work/v1v/license-maven-plugin/target/it/ISSUE-40/target/generated-resources/licenses/common public license version 1.0 - cpl1.0.txt.html file is missing or a directory.
Assertion failed: 

assert false
```

and
``` bash
[INFO] --- license-maven-plugin:1.15-SNAPSHOT:download-licenses (default-cli) @ it-test ---
[WARNING] Unable to retrieve license for dependency: junit:junit
[WARNING] http://www.opensource.org/licenses/cpl1.0.txt
[WARNING] Server returned HTTP response code: 403 for URL: https://opensource.org/licenses/cpl1.0.txt
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 4.052 s
[INFO] Finished at: 2018-01-24T14:52:19Z
[INFO] Final Memory: 16M/215M
[INFO] ------------------------------------------------------------------------
Running post-build script: /Users/victor/work/v1v/license-maven-plugin/target/it/download-licenses-basic/postbuild.groovy
[...]

PR #122, opened 2018-01-25
[121] Fix broken license link in some ITs

v1v:
See https://github.com/mojohaus/license-maven-plugin/issues/121

#### Changes
- JUnit 4.12 uses another [license file](https://github.com/junit-team/junit4/blame/r4.12/pom.xml#L20).
- Fix broken ITs.


tchemit:
good :) the other url is still not reachable

PR #123, opened 2018-01-25
Extended the parameter "missingFileUrl"

Captain-P-Goldfish:
an optional URL parameter that is used as an extension to the missingFile parameter.


tchemit:
thanks for the patch :)

Issue #129, opened 2018-04-14
'remove-file-header' goal leaves the '\n' character, which should be avoided.


xeagle2:
Hello there,

When the RemoveFileHeaderMojo removes the license it calculates the lastChar of '\n'

```
while ( lastchar != '\n' )
{
            lastchar = content.charAt( ( ++lastIndex ) );
}
```

A few lines lower there is a concatenation

`String contentWithoutHeader = content.substring( 0, firstIndex ) + content.substring( lastIndex );`

where **substring** method takes the lastIndex inclusive which is '\n'.

ppalaga:
The fix seems to have been merged. Feel free to reopen if that's not the case.

PR #131, opened 2018-04-14
Fixed ''remove-file-header' goal leaves the '\n' character, which should be avoided. #129'

xeagle2:
Fixed ''remove-file-header' goal leaves the '\n' character, which should be avoided. #129'


tchemit:
thanks for patch

Issue #136, opened 2018-05-26
AbstractAddThirdPartyMojo.consolidate logs warnings for dependencies even if a THIRD-PARTY.properties describes them


srdo:
The checkUnsafeDependencies method logs warnings if the unsafeDependencies set isn't empty https://github.com/mojohaus/license-maven-plugin/blob/70b96dfd6db8b5bde085567ea3cb2ef1bdef2d83/src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java#L876

This is fine and expected when this method is called after missing file files/artifacts have been taken into account, since the plugin should log artifacts with no license. Unfortunately the method is also used in a couple of places in AbstractAddThirdPartyMojo https://github.com/mojohaus/license-maven-plugin/blob/70b96dfd6db8b5bde085567ea3cb2ef1bdef2d83/src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java#L684

This causes the plugin to log warnings for all dependencies, regardless of whether entries exist for them in the missing file. I'm trying to use the plugin to fail the build if there are dependencies with unknown licenses involved, and it would be much more useful if the plugin would only print warnings for the dependencies whose licenses are still unknown after the missing file is resolved.

PR #137, opened 2018-05-26
Issue 136: Don't log warnings about missing licenses until the missin…

srdo:
…g file is taken into account

See the description at https://github.com/mojohaus/license-maven-plugin/issues/136.


tchemit:
thanks for the patch ;)

PR #141, opened 2018-06-06
 add epl_v2 & epl_only_v2

fx-world:
Added the Eclipse Public License - v 2.0


tchemit:
thanks for the patch :)

PR #142, opened 2018-06-12
Add functionality to export third party dependecies in csv format

bennypi:
In order to create a machine readable dependency list that can be imported into e.g. excel, a parameter and an ftl template is added to create a csv file. It contains the groupid, artifactid, version and license(s).

### DIFF
--- a/src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
+++ b/src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
@@ -466,6 +466,12 @@ public abstract class AbstractAddThirdPartyMojo
     boolean sortArtifactByName;
 
     /**
+     * A flag to create the third-party file as a csv file
+     */
+    @Parameter( property = "license.createCSV", defaultValue = "false")
+    boolean createCSV;
+
+    /**
      * Template used to build the third-party file.
      * <p>
      * (This template uses freemarker).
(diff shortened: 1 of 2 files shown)


src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
```diff
@@ -1004,7 +1010,15 @@ void writeThirdPartyFile()
             {
                 licenseMap1 = licenseMap.toLicenseMapOrderByName();
             }
-            thirdPartyTool.writeThirdPartyFile( licenseMap1, thirdPartyFile, isVerbose(), getEncoding(), fileTemplate );
+            if ( createCSV )
+            {
+              getLog().info("Creating CSV file");
+              thirdPartyTool.writeThirdPartyFile( licenseMap1, thirdPartyFile, isVerbose(), getEncoding(), fileTemplate );
```
Jurrie: Is it me, or is the only difference between the `if` and the `else` branch a log statement? Lines 1016 and 1020 seem identical..?

src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
```diff
@@ -465,6 +465,12 @@
     @Parameter( property = "license.sortArtifactByName", defaultValue = "false" )
     boolean sortArtifactByName;
 
+    /**
+     * A flag to create the third-party file as a csv file
```
ppalaga: This comment is not informative enough. It should say clearly what happens if the value is `true` and what happens if the value is `false`. Good style: "If `true` does x, otherwise does y".

BTW, where will the CSV file going to get stored? Should there not be an option for that? Can such a file path option not replace this `createCSV` flag? I mean `csvPath` being null clearly implies `createCSV = false`.

src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
ppalaga: +1

ppalaga:
Closing due to no reaction from @bennypi for 22 days. @bennypi, feel free to re-submit this PR with the above comments resolved.

Issue #46, opened 2016-06-07
Proxy problems when using goal "download-licenses"


Franges:
We got a project that we are using your plugin for aggregating license information and adding the license text files to our delivery artifact within the package phase. After appending the “download-licenses” goal, we got this strange behavior that we can’t deploy to our maven repository anymore.

The build will fail with a 502 proxy error. The problems occurs only using goal “download-licenses”, “aggregate-add-third-party” works just fine. So we went a little deeper and located the problem in the class “AbstractDownloadLicensesMojo.java” inside the initProxy() method. The method sets system property values for the proxy configuration on a global level and overwrites the maven proxy settings.

For now as a workaround, we use the properties-maven-plugin to reset the proxy settings straight after using the license plugin.

PR #146, opened 2018-09-06
HTTP Proxy fixes

jonathanchristison:
Fixes for issue https://github.com/mojohaus/license-maven-plugin/issues/46
Added httpOnly property to hack around proxies that do not support https

Example of httpOnly
```xml
            <plugin>
              <groupId>org.codehaus.mojo</groupId>
              <artifactId>license-maven-plugin</artifactId>
              <version>1.17-SNAPSHOT</version>
              <executions>
                <execution>
                  <goals>
                    <goal>aggregate-download-licenses</goal>
                  </goals>
                  <phase>generate-resources</phase>
                  <configuration>
                    <httpOnly>true</httpOnly>  
                    <licensesOutputDirectory>${license.dir}/licenses</licensesOutputDirectory>
                    <licensesOutputFile>${license.dir}/licenses.xml</licensesOutputFile>
                  </configuration>
                </execution>
              </executions>
            </plugin>
```

### DIFF
--- a/src/main/java/org/codehaus/mojo/license/AbstractDownloadLicensesMojo.java
+++ b/src/main/java/org/codehaus/mojo/license/AbstractDownloadLicensesMojo.java
@@ -219,6 +219,8 @@ public abstract class AbstractDownloadLicensesMojo
     @Parameter( defaultValue = "${project}", readonly = true )
     private MavenProject project;
 
+    @Parameter( property = "httpOnly", defaultValue = "false")
+    private boolean httpOnly;
     // ----------------------------------------------------------------------
     // Plexus Components
     // ----------------------------------------------------------------------
@@ -264,6 +266,18 @@ public abstract class AbstractDownloadLicensesMojo
     protected SortedMap<String, MavenProject> getDependencies( MavenProject project )
     {
         return dependenciesTool.loadProjectDependencies( project, this, localRepository, remoteRepositories, null );
+    }
+
+    protected static java.util.Properties systemProperties;
+    
+    protected void storeProperties()
+    {
+        systemProperties = (java.util.Properties) System.getProperties().clone();
+    }
+    protected void restoreProperties()
+    {
+        if (systemProperties != null)
+            System.setProperties(systemProperties);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jonathanchristison:
Weird seem to pass all ITs with java 7 on my machine

``` bash
java -version 
java version "1.7.0_75"
Java(TM) SE Runtime Environment (build 1.7.0_75-b13)
Java HotSpot(TM) 64-Bit Server VM (build 24.75-b04, mixed mode)
```

```
[INFO] -------------------------------------------------
[INFO] Build Summary:
[INFO]   Passed: 61, Failed: 0, Errors: 0, Skipped: 0
[INFO] -------------------------------------------------
```
Could this just be a glitch in the test env maybe? or is it a classic case of "Works on my machine™" :smile: let me know if it continues to fail and i'll try and track down why

src/main/java/org/codehaus/mojo/license/AbstractDownloadLicensesMojo.java
```diff
@@ -219,6 +219,8 @@
     @Parameter( defaultValue = "${project}", readonly = true )
     private MavenProject project;
 
+    @Parameter( property = "httpOnly", defaultValue = "false")
```
k-wall: I think it would be least surprise for users if property was named `license.httpOnly`.

ppalaga: +1

src/main/java/org/codehaus/mojo/license/AbstractDownloadLicensesMojo.java
```diff
@@ -266,6 +268,18 @@ protected MavenProject getProject()
         return dependenciesTool.loadProjectDependencies( project, this, localRepository, remoteRepositories, null );
     }
 
+    protected static java.util.Properties systemProperties;
```
psakar: why static ?

k-wall:
I made a suggestion #154 that would allow urls to be rewritten or substituted using regexp.  This could be used to achieve the effect of 'httpOnly' suggested by this PR.

ppalaga:
@jonathanchristison could you please rebase and address the previous comments?

jonathanchristison:
@ppalaga I've rebased to latest master, I like @k-wall suggestion for the regex as its dual purpose so I can certainly try and rehunk the commit to only put in the fix for the proxy overwriting issue and leave off   or remap the license.httpOnly feature if we decide to go with that.

ppalaga:
> I like @k-wall suggestion for the regex as its dual purpose so I can certainly try and rehunk the commit to only put in the fix for the proxy overwriting issue and leave off or remap the license.httpOnly feature if we decide to go with that.

+1, please do.

And please squash your commits into one.

Issue #145, opened 2018-08-03
license-maven-plugin interferes with project dependencies


jimonthebarn:
Hey guys,

I ran into an interesting problem I could not solve by myself and was wondering whether you could chime in and help. Its seems that the license-maven-plugin interferes (a dependency of my project even though declared as compile time dependency is skipped entirely) with maven dependencies in certain project setups.

I created a small project portraying the setup and the issue:
https://github.com/jimonthebarn/maven-license-plugin-issue

Could you please take a look?

derKrischan:
I've got the same behaviour. Took me hours to figure out that the license-plugin caused my build to fails because the classes in submodul1 are not found while compiling sumbodule2 as reactor build from parent pom. License plugin is configured to execute aggregate-add-third-party in default phase.

jimonthebarn:
Friendly reminder. :)

jimonthebarn:
@derKrischan Have you found a solution that is not based upon removing the plugin altogether?

derKrischan:
@jimonthebarn : No, I haven't found a solution yet.

tchemit:
Hi guys, I will try to find some time to help you later in the week.
Sorry to be so away from this right now.

PR #148, opened 2018-10-04
Issue 145: Fix aggregate-add-third-party modifying project dependencies

srdo:
Fixes https://github.com/mojohaus/license-maven-plugin/issues/145.

The issue occurs because the plugin modifies the MavenProject dependency lists in DefaultDependencyTool.loadProjectArtifacts. Dependencies may in some cases be removed if they belong to the same reactor as the current project.  

I've added a modified version of the reproduction code from https://github.com/jimonthebarn/maven-license-plugin-issue as a test.

The changes are attempting to preserve the current behavior of the plugin, while also not modifying the MavenProject fields.

### DIFF
--- a/src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
+++ b/src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
@@ -61,6 +61,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.SortedMap;
 import java.util.SortedSet;
+import org.codehaus.mojo.license.api.DependenciesToolException;
 
 /**
  * Abstract mojo for all third-party mojos.
@@ -574,8 +575,9 @@ public abstract class AbstractAddThirdPartyMojo
      * Loads the dependencies of the project (as {@link MavenProject}, indexed by their gav.
      *
      * @return the map of dependencies of the maven project indexed by their gav.
+     * @throws DependenciesToolException if the dependencies could not be loaded
      */
-    protected abstract SortedMap<String, MavenProject> loadDependencies();
+    protected abstract SortedMap<String, MavenProject> loadDependencies() throws DependenciesToolException;
 
     /**
      * Creates the unsafe mapping (says dependencies with no license given by their pom).
@@ -583,12 +585,14 @@ public abstract class AbstractAddThirdPartyMojo
      * Can come from loaded missing file or from dependencies with no license at all.
      *
      * @return the map of usafe mapping indexed by their gav.
-     * @throws ProjectBuildingException if could not create maven porject for some dependencies
+     * @throws ProjectBuildingException if could not create maven project for some dependencies
      * @throws IOException              if could not load missing file
      * @throws ThirdPartyToolException  for third party tool error
+     * @throws DependenciesToolException if the dependencies could not be loaded
      */
     protected abstract SortedProperties createUnsafeMapping()
-      throws ProjectBuildingException, IOException, ThirdPartyToolException, MojoExecutionException;
+      throws ProjectBuildingException, IOException, ThirdPartyToolException,
+            MojoExecutionException, DependenciesToolException;
 
     // ----------------------------------------------------------------------
     // AbstractLicenseMojo Implementaton
@@ -674,7 +678,8 @@ public abstract class AbstractAddThirdPartyMojo
     }
 
     void consolidate() throws IOException, ArtifactNotFoundException, ArtifactResolutionException, MojoFailureException,
-                              ProjectBuildingException, ThirdPartyToolException, MojoExecutionException
+                              ProjectBuildingException, ThirdPartyToolException,
+                              MojoExecutionException, DependenciesToolException
     {
 
         unsafeDependencies = getHelper().getProjectsWithNoLicense( licenseMap );
--- a/src/main/java/org/codehaus/mojo/license/AbstractDownloadLicensesMojo.java
+++ b/src/main/java/org/codehaus/mojo/license/AbstractDownloadLicensesMojo.java
@@ -56,6 +56,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.SortedMap;
+import org.codehaus.mojo.license.api.ResolvedProjectDependencies;
 
 /**
  * Created on 23/05/16.
@@ -285,7 +286,9 @@ public abstract class AbstractDownloadLicensesMojo
 
     protected SortedMap<String, MavenProject> getDependencies( MavenProject project )
     {
-        return dependenciesTool.loadProjectDependencies( project, this, localRepository, remoteRepositories, null );
+        return dependenciesTool.loadProjectDependencies(
+                new ResolvedProjectDependencies( project.getArtifacts(), project.getDependencyArtifacts() ),
+                this, localRepository, remoteRepositories, null );
     }
 
     /**
--- a/src/main/java/org/codehaus/mojo/license/AggregatorAddThirdPartyMojo.java
+++ b/src/main/java/org/codehaus/mojo/license/AggregatorAddThirdPartyMojo.java
@@ -24,14 +24,13 @@ package org.codehaus.mojo.license;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.SortedMap;
 import java.util.SortedSet;
-import java.util.TreeMap;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.maven.artifact.Artifact;
-import org.apache.maven.model.Dependency;
 import org.apache.maven.model.Plugin;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.logging.Log;
@@ -208,14 +207,8 @@ public class AggregatorAddThirdPartyMojo extends AbstractAddThirdPartyMojo
         }
 
         String addThirdPartyRoleHint = groupId + ":" + artifactId + ":" + version + ":" + "add-third-party";
-        Map<String, List<Dependency>> reactorProjectDependencies = new TreeMap<>();
-        for ( MavenProject reactorProject : this.reactorProjects )
-        {
-            reactorProjectDependencies.put( String.format( "%s:%s", reactorProject.getGroupId(),
-                    reactorProject.getArtifactId() ), reactorProject.getDependencies() );
-        }
 
-        for ( Object reactorProject : reactorProjects )
+        for ( MavenProject reactorProject : reactorProjects )
         {
             if ( getProject().equals( reactorProject ) )
             {
@@ -223,11 +216,10 @@ public class AggregatorAddThirdPartyMojo extends AbstractAddThirdPartyMojo
                 continue;
             }
 
-
             AddThirdPartyMojo mojo = (AddThirdPartyMojo) getSession()
                     .lookup( AddThirdPartyMojo.ROLE, addThirdPartyRoleHint );
 
-            mojo.initFromMojo( this, (MavenProject) reactorProject, reactorProjectDependencies );
+            mojo.initFromMojo( this, reactorProject, new ArrayList<>( this.reactorProjects ) );
 
             LicenseMap childLicenseMap = mojo.getLicenseMap();
             if ( isVerbose() )
--- a/pom.xml
+++ b/pom.xml
@@ -365,6 +365,11 @@
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-invoker-plugin</artifactId>
           <version>3.0.0</version>
+          <configuration>
+            <scriptVariables>
+              <projectVersion>${project.version}</projectVersion>
+            </scriptVariables>
+          </configuration>
         </plugin>
         <plugin>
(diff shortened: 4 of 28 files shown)


src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
```diff
@@ -559,6 +561,12 @@
      * @since 1.4
      */
     private Map<String, String> globalKnownLicenses;
+    
+    boolean isAggregatorBuild;
```
ppalaga: Can `isAggregatorBuild` ever be false? Both subclasses set it to `true` or I oversee something? Can perhaps `isAggregatorBuild` be eliminated through encapsulation?

src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
```diff
@@ -559,6 +561,12 @@
...
+    
+    Map<String, List<Dependency>> reactorProjectDependencies;
```
ppalaga: Please explain in JavaDoc what is this field good for.

src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
```diff
@@ -559,6 +561,12 @@
...
+    
+    LoadedProjectDependencies dependencyArtifacts;
```
ppalaga: Please explain in JavaDoc what is this field good for.

src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
```diff
@@ -634,6 +642,12 @@ protected void init()
             // not generating bundled file
             doGenerateBundle = false;
         }
+        
+        if (isAggregatorBuild) {
```
ppalaga: Please apply the code style of the project (if you choose to keep this).

src/main/java/org/codehaus/mojo/license/AbstractDownloadLicensesMojo.java
```diff
@@ -263,7 +264,9 @@ protected MavenProject getProject()
 
     protected SortedMap<String, MavenProject> getDependencies( MavenProject project )
     {
-        return dependenciesTool.loadProjectDependencies( project, this, localRepository, remoteRepositories, null );
+        return dependenciesTool.loadProjectDependencies( 
+                new LoadedProjectDependencies(project.getArtifacts(), project.getDependencyArtifacts()),
```
ppalaga: Code style

src/main/java/org/codehaus/mojo/license/AggregatorAddThirdPartyMojo.java
```diff
@@ -157,6 +157,11 @@ protected void init() throws Exception {
             missingLicensesFileArtifact = aggregateMissingLicensesFileArtifact;
         }
 
+        isAggregatorBuild = true;
+        reactorProjectDependencies = new TreeMap<>();
+        for (MavenProject reactorProject : this.reactorProjects) {
+            reactorProjectDependencies.put(String.format("%s:%s", reactorProject.getGroupId(), reactorProject.getArtifactId()), reactorProject.getDependencies());
```
ppalaga: Code style

src/main/java/org/codehaus/mojo/license/api/LoadedProjectDependencies.java
```diff
@@ -0,0 +1,28 @@
+
```
ppalaga: Missing file header

src/main/java/org/codehaus/mojo/license/api/LoadedProjectDependencies.java
```diff
@@ -0,0 +1,28 @@
...
+package org.codehaus.mojo.license.api;
+
+import java.util.Set;
+import org.apache.maven.artifact.Artifact;
+
+public class LoadedProjectDependencies {
```
ppalaga: Missing JavaDoc

src/main/java/org/codehaus/mojo/license/api/LoadedProjectDependencies.java
```diff
@@ -0,0 +1,28 @@
...
+
+    private final Set<Artifact> allDependencies;
+    private final Set<Artifact> directDependencies;
+
+    public LoadedProjectDependencies( Set<Artifact> allDependencies, Set<Artifact> directDependencies )
```
ppalaga: I vote for making this class immutable. Either copy the constructor params and wrap them in an Collections.unmodifiableSet or find another way to reach the same effect.

src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
srdo: It is false when using `AddThirdPartyPlugin` as a goal. It is only true when the plugin is called via `AggregateAddThirdPartyMojo`. The line in `AddThirdPartyPlugin` where it is set to true is only called through `AggregateAddThirdPartyMojo` (it essentially invokes the plugin once for each child project in the reactor).

Regarding whether it can be eliminated, I don't think so without a larger refactoring. The problem is that the `init` method needs to behave differently when used as part of an `AggregateAddThirdPartyMojo` invocation, both when called directly on the `AggregateAddThirdPartyMojo` class, and the `AddThirdPartyMojo` instances it creates. `init` doesn't take parameters, and the interface is defined at the `AbstractLicenseMojo` level. I didn't want to start rewriting these classes to not use inheritance so heavily, so adding a field instead was the only solution I could think of.

src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
srdo: Will do

src/main/java/org/codehaus/mojo/license/api/LoadedProjectDependencies.java
srdo: Will do.

src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
ppalaga: > It is false when using AddThirdPartyPlugin as a goal.

I see `isAggregatorBuild = true` in `AddThirdPartyPlugin` https://github.com/mojohaus/license-maven-plugin/pull/148/files#diff-f71923bddd2327bb3c7b3c62c80878c2R467

> Regarding whether it can be eliminated, I don't think so without a larger refactoring. The problem is that the init method needs to behave differently when used as part of an AggregateAddThirdPartyMojo invocation, both when called directly on the AggregateAddThirdPartyMojo class, and the AddThirdPartyMojo instances it creates. init doesn't take parameters, and the interface is defined at the AbstractLicenseMojo level.

Not sure I understand all details, but can't you define something like `protected abstract LoadedProjectDependencies loadDependencyArtifacts()` in `AbstractAddThirdPartyMojo`, implement it differently in the two subclasses and replace this conditional https://github.com/mojohaus/license-maven-plugin/pull/148/files#diff-0bfb2b7e70b9abacb4339555c7b49e81R668 with the follwing? 

```
dependencyArtifacts = loadDependencyArtifacts()
```

ppalaga: BTW, if you really intended to have isAggregatorBuild = false in AddThirdPartyPlugin https://github.com/mojohaus/license-maven-plugin/pull/148/files#diff-f71923bddd2327bb3c7b3c62c80878c2R467 and all tests are passing, then there is probably a test missing that would fail due to this mistake?

srdo: > I see isAggregatorBuild = true in AddThirdPartyPlugin https://github.com/mojohaus/license-maven-plugin/pull/148/files#diff-f71923bddd2327bb3c7b3c62c80878c2R467

Yes. This line is in the `initFromMojo` method, which is used by `AggregatorAddThirdPartyMojo` to initialize the `AddThirdPartyMojo` instances it creates (here: https://github.com/mojohaus/license-maven-plugin/pull/148/files#diff-38b03aa7a6c189ea0f7e814313a14682R215). It isn't used when running `AddThirdPartyMojo` as a goal directly.

This is also why doing something like `protected abstract LoadedProjectDependencies loadDependencyArtifacts()` doesn't work. The behavior of `loadDependencyArtifacts()` would need to do one thing when `AddThirdPartyMojo` runs it as part of the `addThirdParty` goal, while it needs to do something else when `AggregatorAddThirdPartyMojo` instantiates a bunch of `AddThirdPartyMojo`s to run for its child projects.

Trying to give an overview:
* Running the `addThirdParty` goal from Maven: `init` has to act as if `isAggregatorBuild == false`. 
* Running `aggregateAddThirdParty` will run `init` once through `AggregatorAddThirdPartyMojo` for the parent project, where `isAggregatorBuild == true`, and then run `init` once per child project through `AddThirdPartyMojo.initFromMojo` where `isAggregatorBuild == true`.

src/main/java/org/codehaus/mojo/license/api/LoadedProjectDependencies.java
```diff
@@ -0,0 +1,63 @@
...
+    {
+        this.allDependencies = new HashSet<>( allDependencies );
+        this.directDependencies = new HashSet<>( directDependencies );
+    }
+
+    public Set<Artifact> getAllDependencies()
+    {
+        return Collections.unmodifiableSet( allDependencies );
```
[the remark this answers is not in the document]
srdo: Ok, will change

srdo:
I think there's a small bug in `DefaultDependenciesTool.loadProjectDependencies`, will try to make a test that covers it and update.

src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
ppalaga: > Yes. This line is in the `initFromMojo` method, which is used by `AggregatorAddThirdPartyMojo` to initialize the `AddThirdPartyMojo` instances it creates (here: https://github.com/mojohaus/license-maven-plugin/pull/148/files#diff-38b03aa7a6c189ea0f7e814313a14682R215). It isn't used when running `AddThirdPartyMojo` as a goal directly.

OK, thanks for the explanation, I was not aware of the fact that `AggregatorAddThirdPartyMojo`  instantiates `AddThirdPartyMojo`s.

Given what you say, `isAggregatorBuild` seems to be significant for `AddThirdPartyMojo` only (because it is always true for `AggregatorAddThirdPartyMojo`). Could we perhaps move `isAggregatorBuild` to `AddThirdPartyMojo` and introduce a `protected abstract LoadedProjectDependencies loadDependencyArtifacts()` in `AbstractAddThirdPartyMojo` and let  it alwas return

* `dependenciesTool.loadProjectArtifacts( localRepository, remoteRepositories, project, reactorProjectDependencies )` from `AggregatorAddThirdPartyMojo` and
* the current conditional from `AddThirdPartyMojo`? 

I think this would improve the readability of the code.

srdo: Made this change. Turns out the method wasn't necessary in the aggregator mojo, so I removed it from there.

srdo:
Rebased, and fixed another few bugs in `loadProjectArtifacts`. I wound up replacing the method. The previous implementation would fail to find transitive dependencies for long chains of sibling dependencies in the same reactor, and would use the wrong list of remote repositories for looking up dependencies. Both issues are demonstrated in it/ISSUE-145-2. The latter bug would cause the build to crash. 

The issue with wrong transitive dependencies wasn't a big deal, but I think we might as well fix it. When the aggregator mojo finds dependencies for sibling projects, it was possible to be missing some dependencies in the return from `loadProjectArtifacts`. For instance if you have siblings A, B, C and A depends on B depends on C, then `loadProjectArtifacts` for A would replace B with C in the direct dependency list. This was wrong, but since the aggregator mojo processes all modules in the reactor, B would still be picked up when the aggregator mojo moved on to processing B.

srdo:
The Java 7 tests failed, they pass for me locally. Going to retrigger the build.

ppalaga:
> Rebased, and fixed another few bugs in `loadProjectArtifacts`. I wound up replacing the method. The previous implementation would fail to find transitive dependencies for long chains of sibling dependencies in the same reactor, and would use the wrong list of remote repositories for looking up dependencies. Both issues are demonstrated in it/ISSUE-145-2. 

Thanks a lot for the fixes and special thanks for the test!

> The issue with wrong transitive dependencies wasn't a big deal, but I think we might as well fix it.

Not sure you imply (a) you fixed it or (b) it needs to be fixed. If (b) is the case, please file a new issue.

The PR looks good to me now. Please squash the commits, and reword to something like `Fix #145 aggregate-add-third-party modifying project dependencies`  and I'll merge afterwards.

srdo:
> Not sure you imply (a) you fixed it or (b) it needs to be fixed. If (b) is the case, please file a new issue.

It is fixed now. 

Squashed.

ppalaga:
@srdo good work, thanks!

PR #150, opened 2018-10-05
Fix plurals in messages emitted for dependencies without license

### LINKED ISSUE #149: Incorrect warning message reporting count of dependencies with no license
1. If there is only 1 dependency with no license, the message is:

```
[WARNING] There is 1 dependencies with no license
```

where it should be:

```
[WARNING] There is 1 dependency with no license
```

2. If there are multiple dependencies with no license, the message is:

```
[WARNING] There is 2 dependencies with no license
```

where it should be:

```
[WARNING] There are 2 dependencies with no license
```

dpursehouse:
Fixes #149

Change-Id: Ie46be7830280f55778123213c16095d938cc4042
Signed-off-by: David Pursehouse <dpursehouse@collab.net>

### DIFF
--- a/src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
+++ b/src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
@@ -879,12 +879,20 @@ public abstract class AbstractAddThirdPartyMojo
         if ( CollectionUtils.isNotEmpty( unsafeDeps ) )
         {
             Log log = getLog();
-            log.warn( "There is " + unsafeDeps.size() + " dependencies with no license :" );
-            for ( MavenProject dep : unsafeDeps )
+            if ( log.isWarnEnabled() )
             {
+                boolean plural = unsafeDeps.size() > 1;
+                String message = String.format( "There %s %d %s with no license :",
+                    plural ? "are" : "is",
+                    unsafeDeps.size(),
+                    plural ? "dependencies" : "dependency" );
+                log.warn( message );
+                for ( MavenProject dep : unsafeDeps )
+                {
 
-                // no license found for the dependency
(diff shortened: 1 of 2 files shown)


src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
```diff
@@ -879,12 +879,20 @@ void checkUnsafeDependencies()
         if ( CollectionUtils.isNotEmpty( unsafeDeps ) )
         {
             Log log = getLog();
-            log.warn( "There is " + unsafeDeps.size() + " dependencies with no license :" );
-            for ( MavenProject dep : unsafeDeps )
+            if ( log.isWarnEnabled() )
             {
+                boolean plural = unsafeDeps.size() > 1;
```
ppalaga: If `unsafeDeps.size() == 0` the message will be "There is 0 dependency...". Clearly, "There are 0 dependencies..." would be better. Could you please change the condition accordingly?

src/main/java/org/codehaus/mojo/license/api/DefaultThirdPartyTool.java
```diff
@@ -175,12 +175,20 @@ public void attachThirdPartyDescriptor( MavenProject project, File file )
...
             else
             {
-                log.debug( "There is " + unsafeDependencies.size() + " dependencies with no license from poms : " );
-                for ( MavenProject dep : unsafeDependencies )
+                if ( log.isDebugEnabled() )
                 {
-
-                    // no license found for the dependency
-                    log.debug( " - " + MojoHelper.getArtifactId( dep.getArtifact() ) );
+                    boolean plural = unsafeDependencies.size() > 1;
```
ppalaga: Same as above: change the condition so that zero is handled as plural.

ppalaga:
Thanks for the contribution, @dpursehouse!

Issue #143, opened 2018-06-22
enhance overrideFile with overrideFileUrl


TorstenKruse:
Like the missing file it would be nice to maintain a company wide file with overrides which could be references via an url. Then it would not be neccessary to add the same overrides to every single project.

Jurrie:
Same goes for licenseMergesFile.

Jurrie:
This request is implemented in https://github.com/mojohaus/license-maven-plugin/pull/153.

PR #152, opened 2018-10-12
Property licenseMergesUrl is not in 1.16; it's not yet released.

Jurrie:
https://github.com/mojohaus/license-maven-plugin/issues/143 is not yet released. So the `@since` annotation is wrong.


ppalaga:
Thanks @Jurrie !

PR #153, opened 2018-10-13
Enhance overrideFile with overrideUrl - Closes #143

### LINKED ISSUE #143: enhance overrideFile with overrideFileUrl

Jurrie:
This is basically 545ba8a8033666829763be77e305461e77a47180 (which was
for licenseMergesFile -> licenseMergesUrl) applied to overrideFile ->
overrideUrl.

The downside is that we can not support a default value for overrideFile
anymore. So users upgrading should explicitly set the overrideUrl
configuration property.


ppalaga:
Replaced by https://github.com/mojohaus/license-maven-plugin/pull/181 closing this one.

PR #156, opened 2018-11-07
Extends documentation for include/exclude scopes

Weltraumschaf:
The problem was that the property names are in plural (speaking from
included/ecluded scopes). So as a user I assume that I can add multiple
scopes, but nowhere in the documentation the syntax for that is mentioned.

After digging through the source code I found that the scopes are delimitted
by comma.

So I extend the documentation for the properties

- license.excludedScopes
- license.includedScopes

how to add multiple scopes.

### DIFF
--- a/src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
+++ b/src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
@@ -116,6 +116,9 @@ public abstract class AbstractAddThirdPartyMojo
 
     /**
      * A filter to exclude some scopes.
+     * <p>
+     * Multiple scopes are delimited by comma, eg. {@code <excludedScopes>test,provided</excludedScopes>}
+     * or {@code -Dlicense.excludedScopes="test,provided"}.
      *
      * @since 1.1
      */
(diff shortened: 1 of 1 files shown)


src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
```diff
@@ -116,6 +116,9 @@
 
     /**
      * A filter to exclude some scopes.
+     * <p>
+     * Multiple scopes are delimited by comma, eg. {@code <excludedScopes>test,provided</excludedScopes>}
+     * or {@code -Dlicense.excludedScopes="test,provided"}.
```
ppalaga: `-Dlicense.excludedScopes="test,provided"` perhaps works, but the quotes are not necessary as long as there is no whitespace there. It works without quotes on all platforms. So please remove the quotes.

src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
```diff
@@ -124,6 +127,9 @@
 
     /**
      * A filter to include only some scopes, if let empty then all scopes will be used (no filter).
+     * <p>
+     * Multiple scopes are delimited by comma, eg. {@code <includedScopes>test,provided</includedScopes>}
+     * or {@code -Dlicense.includedScopes="test,provided"}.
```
ppalaga: Same as above: please remove the quotes.

ppalaga:
Changes adopted in https://github.com/mojohaus/license-maven-plugin/pull/180 , closing this one. Thanks @Weltraumschaf!

Weltraumschaf:
Sorry, your comment was undergone in spam...

PR #157, opened 2018-11-16
Ignore Eclipse project files

ppalaga:
@tchemit could you please review?


ppalaga:
The build is passing now. @tchemit could you please review?

ppalaga:
I removed the fix for #158, keeping just the maintenance commits.

Issue #159, opened 2018-11-28
Listing the mainLicense multiple times in licenseMerges causes the plugin to act as if the license is not present.


srdo:
When using the licenseMerges option in addThirdParty or aggregateAddThirdParty, there's a check that the user doesn't list the same license twice, except for main licenses. I think the check should also check main licenses if they appear in the same licenseMerge list.

For example:
```
<licenseMerge>
Common Development and Distribution License (CDDL) v1.1 | 
COMMON DEVELOPMENT AND DISTRIBUTION LICENSE (CDDL) Version 1.1 |
Common Development and Distribution License (CDDL) v1.1
</licenseMerge>
```

This will currently cause the plugin to act as if the license isn't present on the dependency, i.e. the license will be missing from THIRD-PARTY.txt.

I think it would be nice if the existing check for duplicates also checked that each licenseMerge element only contains the main license once.

PR #160, opened 2018-11-28
Issue 159: Check that the main license only appears once per licenseM…

srdo:
…erge element

Closes https://github.com/mojohaus/license-maven-plugin/issues/159


ppalaga:
Thanks, @srdo !

Issue #161, opened 2018-11-29
Allow users to specify globalKnownLicenses in pom.xml


ptahchiev:
I know unknown licenses can be added with the `missingFile` configuration property, but this is very cumbersome when you have a multi-module project. In my case I have a project with 40 sub-modules and the license-plugin is declared in the top project pom.xml, however if I am asked to add `missingFile`to all the submodules. I would like to specify the missing licenses in the configuration of the plugin, such as:
````
                            <globalKnownLicenses>
                                <dom4j--dom4j--1.6.1>The Apache Software License, Version 2.0</dom4j--dom4j--1.6.1>
                                <jgridshift--jgridshift--1.0>The Apache Software License, Version 2.0</jgridshift--jgridshift--1.0>
                                <net.jcip--jcip-annotations--1.0>The Apache Software License, Version 2.0</net.jcip--jcip-annotations--1.0>
                                <org.codehaus.jettison--jettison--1.2>The Apache Software License, Version 2.0</org.codehaus.jettison--jettison--1.2>
                                <javax.el--el-api--2.2>Common Development and Distribution License (CDDL)</javax.el--el-api--2.2>
                                <javax.servlet--jstl--1.2>Common Development and Distribution License (CDDL)</javax.servlet--jstl--1.2>
                                <javax.xml.stream--stax-api--1.0-2>Common Development and Distribution License (CDDL)</javax.xml.stream--stax-api--1.0-2>
                                <javax.transaction--jta--1.1>Common Development and Distribution License (CDDL)</javax.transaction--jta--1.1>
                                <com.github--vfs-s3--2.4.2>The Apache Software License, Version 2.0</com.github--vfs-s3--2.4.2>
                                <org.springframework.social--spring-social-twitter--2.0.2.NEMESIS-RELEASE>The Apache Software License, Version 2.0</org.springframework.social--spring-social-twitter--2.0.2.NEMESIS-RELEASE>
                                <org.springframework.social--spring-social-twitter-autoconfigure--2.0.2.NEMESIS-RELEASE>The Apache Software License, Version 2.0</org.springframework.social--spring-social-twitter-autoconfigure--2.0.2.NEMESIS-RELEASE>
                                <com.nemesis.rules--nemesis-rules-targeting--2.0.0.RELEASE>Nemesis Software Commercial License</com.nemesis.rules--nemesis-rules-targeting--2.0.0.RELEASE>
                            </globalKnownLicenses>
````
[...]

PR #162, opened 2018-11-29
161: Allow specifying missing licenses in the configuration of the pl…

ptahchiev:
…ugin

 - use the globalKnownLicenses property to allow customers to declare known licenses.
 - resolveUnsafeDependenciesFromFile method incorrectly checks if missingLicenses file exists and only then reads missingFileUrl property. To fix this we extract the logic in a new method called resolveUnsafeDependencies and in there we resolve dependencies 1) from the dependencies provided in the plugin configuration (globalKnownLicenses) 2) if missingLicenses file is present then we use the content of the file 3) missingFileUrl if the property is actually a url.
 - the doGenerateMissing logic was wrong as the field was initialized before the consolidate method, which is the one that actually resolves the unsafe dependencies. That's why the logic of the doGenerateMissing is now after the consolidate method.

See #161

### DIFF
--- a/src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
+++ b/src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
@@ -502,6 +502,15 @@ public abstract class AbstractAddThirdPartyMojo
     @Parameter( property = "project.artifacts", required = true, readonly = true )
     Set<Artifact> dependencies;
 
+    /**
+     * Map from G/A/V as string to license key, obtained from global dependencies of type=.ld.properties.
+     * This could probably be refactored to have more in common with the classifier-based loader.
+     *
+     * @since 1.4
+     */
+    @Parameter( property = "license.globalKnownLicenses")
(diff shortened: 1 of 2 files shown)


ptahchiev:
@ppalaga I don't think these are separate issues - I had to make three changes because otherwise I could't implement what I wanted.

ppalaga:
> resolveUnsafeDependenciesFromFile method incorrectly checks if missingLicenses file exists and only then reads missingFileUrl property

Sounds like a bug that exists (and is reproducible) even before adding your enhancement with missing licenses in the configuration. 

The same holds for 

> doGenerateMissing logic was wrong as the field was initialized before the consolidate method

So please check if those issues still exist. Maybe they don't because there were some things fixed in this area. But if they still exist, please create separate issues for them, create integration tests for each of them, fix them and send them as separate PRs. I'll review the present PR only after these two separate issues are resolved.

ppalaga:
@ptahchiev could you please rebase this PR on top of current master? I think the two unrelated changes were in between resolved by other contributors.

ppalaga:
Closing a out of date

PR #165, opened 2018-12-14
Issue 164: Fix dependency map sorting by name in add-third-party goal

dzolo:
The fix is done by carrying information whether the License Map was sorted by name.
If so the Third Party Tool sorts the `dependencyMap`.

Additionally sorting is case insensitive which makes more sense for Maven project names.


ppalaga:
@dzolo could you please add an integration test in https://github.com/mojohaus/license-maven-plugin/tree/master/src/it that fails before and passes after applying your fix?

ppalaga:
Replaced by https://github.com/mojohaus/license-maven-plugin/pull/187

PR #183, opened 2019-01-04
Create integration test for issue 174

rez-:


ppalaga:
Please use `git rebase upstream/master` (where master is https://github.com/mojohaus/license-maven-plugin.git ) rather than `git merge`. When you rebase the merge commits disappear.

Second, please amend  the follwing change to your first commit: You need to manage the the plugin in the pom of the test using @pom.version@ like in this other test: https://github.com/mojohaus/license-maven-plugin/blob/0d75b07605c2d660098681511d007c87ef0c7092/src/it/add-third-party-with-deps-name-order/pom.xml#L58-L70

Otherwise the test is run against the latest released version, rather than the version built from source.

rez-:
Thanks for letting me know! I've implemented your comments in a new branch: rez-/issue-174.1

ppalaga:
> I've implemented your comments in a new branch: rez-/issue-174.1

You did not need to. You could re-use your existing branch and force push via `git push --force issue-174 origin/issue-174.1` where `origin` is your git alias of your github fork. That would effectively replace all commits in this PR with the new ones.

rez-:
Hmm when I do that, I get the following message:
fatal: 'issue-174' does not appear to be a git repository
fatal: Could not read from remote repository

ppalaga:
Sorry, I messed the command, `git push --force origin issue-174` should work. Please close either this one or https://github.com/mojohaus/license-maven-plugin/pull/190

PR #185, opened 2019-01-04
Fix basic auth on proxy and avoid overwriting system properties

jonathanchristison:
See https://github.com/mojohaus/license-maven-plugin/pull/146#issuecomment-450842609

Rehunked (should now be only system settings reset and proxy auth fixes) , Rebased and fixed conflicts, its a bit hard to verify the diff after all that but i think its good to go :smile:

### DIFF
--- a/src/main/java/org/codehaus/mojo/license/AbstractDownloadLicensesMojo.java
+++ b/src/main/java/org/codehaus/mojo/license/AbstractDownloadLicensesMojo.java
@@ -291,6 +291,18 @@ public abstract class AbstractDownloadLicensesMojo
                 this, localRepository, remoteRepositories, null );
     }
 
+    protected java.util.Properties systemProperties;
+
+    protected void storeProperties()
+    {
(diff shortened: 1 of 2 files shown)


src/main/java/org/codehaus/mojo/license/AbstractDownloadLicensesMojo.java
```diff
@@ -291,6 +291,18 @@ protected MavenProject getProject()
...
+    protected java.util.Properties systemProperties;
+
+    protected void storeProperties()
+    {
+        systemProperties = (java.util.Properties) System.getProperties().clone();
+    }
+    protected void restoreProperties()
+    {
+        if (systemProperties != null)
+            System.setProperties(systemProperties);
```
ppalaga: Would you please set `systemProperties = null` (outside the if scope)?

ppalaga:
Replaced by https://github.com/mojohaus/license-maven-plugin/pull/188

PR #189, opened 2019-01-08
Fix #135 Content wrongly truncated when adding header after package

### LINKED ISSUE #135: Content wrongly truncated when adding header after package declaration
Hello

I have a java file that has a "package" and "import" on the same line, something like this:

`package org.example.test; import java.io.File;`

When the file is processed by the "update-file-header" goal, the content after the package declaration (the import) is wrongly truncated, and I obtain something like this (that doesn't compile any longer):
```java
package org.example.test;

/*-
 * #%L
 * test.license.header
 * %%
 * Copyright (C) 2018 org.example
 * %%
 * This program is free software: you can redistribute it and/or modify
 * it under the terms of the GNU Lesser General Public License as
 * published by the Free Software Foundation, either version 3 of the
 * License, or (at your option) any later version.
 * 
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU General Lesser Public License for more details.
 * 
 * You should have received a copy of the GNU General Lesser Public
 * License along with this program.  If not, see
 * <http://www.gnu.org/licenses/lgpl-3.0.html>.
 * #L%
 */
mport java.io.File;
```

This is a problem to me because the java file is not hand-written, it is generated by a tool that doesn't allow me to control the generated format .. so I can not really workaround the problem

I have pinpointed the problem to this code on class **JavaFileHeaderTransformer**, 
```java
        String prolog = null;
        int startProlog = content.indexOf( "package" );
        if ( startProlog > -1 )
        {

            // package was detected
            int endProlog = content.indexOf( ";", startProlog );

            if ( endProlog > -1 )
            {

                // prolog end was detected

                int lastIndex = endProlog + 1;

                prolog = content.substring( 0, lastIndex );

                // prolog goes to next line
                prolog += getLineSeparator();

                if ( lastIndex == content.length() )
                {

                    // adding a new empty end line to the content
                    content += getLineSeparator();
                }


            }
        }

        if ( prolog == null )
        {

            // no prolog detected
            result = super.addHeader( header, content );
        }
        else
        {

            // prolog detected
            content = content.substring( prolog.length() );
            result = super.addHeader( prolog + getLineSeparator() + header, content );
        }
        return result;
    }
```
On windows, the line `prolog += getLineSeparator();`  adds two characters (CR+LF) to the prolog, so when it later does `content = content.substring( prolog.length() );` one character is lost from the original content after the "package" declaration.

I attach a project with a single java file that shows the problem (the source directory contains the wrongly truncated java file and the original file before header insertion )
[test.license.header.zip](https://github.com/mojohaus/license-maven-plugin/files/2008210/test.license.header.zip)

Thank you

gevegab:


gevegab:
The test only failed on Windows.  Committed small change to make it reproducible on Linux

gevegab:
The fix is working with jdk8, both in windows and linux, but it fails with jdk11.

I will install jdk11 and test locally in my development machine

ppalaga:
@gevegab could you please squash and rebase now?

gevegab:
Finally integrated, hallelujah !!

ppalaga:
Thanks, @gevegab !

PR #190, opened 2019-01-09
Issue 174

rez-:


src/main/java/org/codehaus/mojo/license/api/DefaultDependenciesTool.java
```diff
@@ -257,6 +257,23 @@
...
+                    if ( !depMavenProject.getArtifactId().equals( artifact.getArtifactId() ) )
+                    {
+                        depMavenProject.setArtifactId( artifact.getArtifactId() );
+                        depMavenProject.getArtifact().setArtifactId( artifact.getArtifactId() );
+                    }
+                    if ( !depMavenProject.getVersion().equals( artifact.getVersion() ) )
+                    {
+                        depMavenProject.setVersion( artifact.getVersion() );
+                        depMavenProject.getArtifact().setVersion( artifact.getVersion() );
+                    }
```
ppalaga: I am not particularly happy about setting fields returned by a Maven component. It smells after side effects, like we recently fixed in https://github.com/mojohaus/license-maven-plugin/issues/145 . But let's try, maybe it will work.

ppalaga:
Thanks, @rez-

PR #191, opened 2019-01-09
Fix #192 Dependency map sorting by project name with ignored case in add-third-party goal

### LINKED ISSUE #192: add-third-party sorts dependency map by project name without case ignore
The **add-third-party** performs ordering of dependencies using _String#compareTo_. This causes that third party file is not properly ordered by alphabet. E.g. _antlr_ dependency is added to the third party file after _Apache Commons Lang_ dependency.

dzolo:
Currently the **add-third-party** performs ordering of dependencies using **String#compareTo**. This causes that e.g. _antlr_ dependency is added to the third party file after _Apache Commons Lang_ dependency. This merge request change this behavior and performs ordering with ignored case.

Integration test add-third-party-with-deps-name-order was extended for dependency ignore case order check.

### DIFF
--- a/src/it/add-third-party-with-deps-name-order/pom.xml
+++ b/src/it/add-third-party-with-deps-name-order/pom.xml
@@ -54,6 +54,11 @@
       <artifactId>commons-lang3</artifactId>
       <version>3.8.1</version>
     </dependency>
+    <dependency>
+      <groupId>antlr</groupId>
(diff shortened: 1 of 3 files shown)


dzolo:
@ppalaga Sorry, fixed.

ppalaga:
Thanks, @dzolo

Issue #208, opened 2019-01-14
ISSUE-154 integration test fails on Windows


gevegab:
The problem is very similar to issue #204, the url of the replacement file specified in the POM of the integration test is wrongly interpolated in windows.

Instead of 
```xml
<replacement>file://${project.basedir}/src/alternative.txt</replacement>
```

Should be
```xml
<replacement>${project.baseUri}src/alternative.txt</replacement>
```

will create PR with the fix

PR #210, opened 2019-01-15
Fix #205 : invalid hashbang handling for shell files  on windows

### LINKED ISSUE #205: Bad hashbang handling on windows
See https://ci.appveyor.com/project/ppalaga/license-maven-plugin-g0x55/builds/21547120/job/k2lp03sc6jror7lr

gevegab:


gevegab:
The integration test is working in my development windows machine, but it is failing on appveyor CI.

I will commit a temporarily modified POM, to have a verbose execution of integration tests, and try to diagnose the problem

gevegab:
Now the windows CI  seems to be working (without any additional change). the integration test for update-file-header-test-mojo is successful. 

The only integration test failing on windows  is ISSUE-154, this is reported in #208 and fixed in #209

src/main/java/org/codehaus/mojo/license/header/transformer/PropertiesFileHeaderTransformer.java
```diff
@@ -67,13 +67,13 @@ public String addHeader( String header, String content )
 
             // shebang was detected
 
-            int endProlog = content.indexOf( getLineSeparator() );
+            int endProlog = content.indexOf( '\n', startProlog );
```
ppalaga: I think the fix is wrongly assuming that all possible shell scripts are delimited by `\n`. I vote for being as defensive as possible here - i.e. we should accept both an `\r\n` pair or `\n` which ever of them is present. Something like you did here: https://github.com/mojohaus/license-maven-plugin/pull/189/files#diff-0dbbbe109835879abe30a4c74cf1e191R131 WDYT?

gevegab: The fix considers both delimiters (`\r\n` and `\n`) : the variable endProlog will point to the end of the line no matter what convention is used.

I tested both cases in my machine, but couldn't commit the integration test with the `\r\n` scenario because the editorconfig verification failed (it is expecting all the plugin code, including tests, to have  LF end of lines)

https://github.com/mojohaus/license-maven-plugin/blob/511684ec861fa0b84f5075527154c50657b99b99/.editorconfig#L8 

and I didn't think was a good idea to modify the main POM just to make an exception for this integration test

ppalaga: > couldn't commit the integration test with the `\r\n` scenario because the editorconfig verification failed (it is expecting all the plugin code, including tests, to have LF end of lines)

Just add a new rule at the end of the top level `.editorconfig` file, something like

```
[src/it/my-test/my-file.ext]
end_of_line = crlf 
```

or you could even create a new `.editorconfig` file in the root folder of your test and add an appropriate rule there.

ppalaga: > The fix considers both delimiters (`\r\n` and `\n`) : the variable endProlog will point to the end of the line no matter what convention is used.

Indeed, thanks for correcting me.

ppalaga:
Thanks, please add the integration test if you have one, and reword the commit to `Fix #205 Bad hashbang handling on Windows`.

gevegab:
I made the requested changes but the build failed on linux with jdk7 (for some unrelated integration tests), can you verify the logs?

ppalaga:
@gevegab I am hereby giving you a virtual medal for helping to make the Windows build pass :-)

Issue #134, opened 2018-04-19
Can't use any url in missingFileUrl


tchemit:
The actual code is:
```
     if (HttpRequester.isStringUrl(missingFileUrl))
            {
                String httpRequestResult = HttpRequester.getFromUrl(missingFileUrl);
                unsafeMappings.load(new ByteArrayInputStream(httpRequestResult.getBytes()));
            }
```
which fails if you use for example a file url.

ppalaga:
Which mojos are affected?

nfalco79:
> Which mojos are affected?

All mojos with missingFileUrl attribute are affected by this:
- add-third-party
- aggregate-add-third-party
- aggregate-third-party-report
- third-party-report

PR #246, opened 2019-02-08
Fix #245 add missingFileUrl in initFromMojo method

### LINKED ISSUE #245: aggregate-add-third-party does not take care the missingFileUrl parameter
I had set the missingFileUrl to execute a aggregate-add-third-party goal but it does not perform any HTTP request. Looking into the code the issue is AddThirdPartyMojo.initFromMojo forget to copy the missingFileUrl attribute.

I had also notice that there is a circular dependency between AggregatorAddThirdPartyMojo -> AddThirdPartyMojo -> AggregatorAddThirdPartyMojo

nfalco79:
Add missingFileUrl in AddThirdPartyMojo initFromMojo method when copy attributes from caller mojo AggregatorAddThirdPartyMojo.

I had also change the signature from AggregatorAddThirdPartyMojo to AbstractAddThirdPartyMojo to remove the cyclic dependency.

### DIFF
--- a/src/it/aggregate-add-third-party-missing-file/invoker.properties
+++ b/src/it/aggregate-add-third-party-missing-file/invoker.properties
@@ -1,0 +1,23 @@
+###
+# #%L
+# License Maven Plugin
+# %%
+# Copyright (C) 2019 - Nikolas Falco
+# %%
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Lesser Public License for more details.
+#
+# You should have received a copy of the GNU General Lesser Public
+# License along with this program.  If not, see
+# <http://www.gnu.org/licenses/lgpl-3.0.html>.
+# #L%
+###
+invoker.goals=clean license:aggregate-add-third-party
+invoker.failureBehavior=fail-never
--- a/src/it/aggregate-add-third-party-missing-file/pom.xml
+++ b/src/it/aggregate-add-third-party-missing-file/pom.xml
@@ -1,0 +1,62 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  #%L
+  License Maven Plugin
+  %%
+  Copyright (C) 2019 - Nikolas Falco
+  %%
+  This program is free software: you can redistribute it and/or modify
+  it under the terms of the GNU Lesser General Public License as
+  published by the Free Software Foundation, either version 3 of the
+  License, or (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Lesser Public License for more details.
+
+  You should have received a copy of the GNU General Lesser Public
+  License along with this program.  If not, see
+  <http://www.gnu.org/licenses/lgpl-3.0.html>.
+  #L%
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>org.codehaus.mojo.license.test</groupId>
+  <artifactId>aggregate-add-third-party-missing-file</artifactId>
+  <version>@pom.version@</version>
+
+  <modules>
+    <module>child1</module>
+  </modules>
+
+  <name>License Test :: aggregate-add-third-party-missing-file</name>
+
+  <packaging>pom</packaging>
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <license.verbose>true</license.verbose>
+    <license.useMissingFile>true</license.useMissingFile>
+    <license.missingFileUrl>http://this.url.does.not.exists/THIRD-PARTY.properties</license.missingFileUrl>
+  </properties>
+
+  <build>
+
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.codehaus.mojo</groupId>
+          <artifactId>license-maven-plugin</artifactId>
+          <version>@pom.version@</version>
+        </plugin>
+      </plugins>
+    </pluginManagement>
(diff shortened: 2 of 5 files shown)


src/main/java/org/codehaus/mojo/license/AddThirdPartyMojo.java
```diff
@@ -469,7 +469,7 @@ private void writeMissingFile()
         }
     }
 
-    void initFromMojo( AggregatorAddThirdPartyMojo mojo, MavenProject mavenProject,
+    void initFromMojo( AbstractAddThirdPartyMojo mojo, MavenProject mavenProject,
```
ppalaga: Is this change necessary to fix the issue? IIRC, AggregatorAddThirdPartyMojo uses (and instantiates) AbstractAddThirdPartyMojo to do the job for the submodules. So this method makes sense only with AggregatorAddThirdPartyMojo arg. I might be wrong, I am not an expert in these mojos.

nfalco79: No, it is not needed. But
> AggregatorAddThirdPartyMojo uses (and instantiates) AbstractAddThirdPartyMojo to do the job for the submodules

the AggregatorAddThirdPartyMojo uses and instantiates the AddThirdPartyMojo to do the job
```
AddThirdPartyMojo mojo = (AddThirdPartyMojo) getSession()
                    .lookup( AddThirdPartyMojo.ROLE, addThirdPartyRoleHint );
mojo.initFromMojo( this, reactorProject, new ArrayList<>( this.reactorProjects ) );
``` 
and AddThirdPartyMojo takes exactly the AggregatorAddThirdPartyMojo. This sound as a circulas dependencies. Since all the task is performed for each module in the reactor it's like run add-third-party goal directly. Anyway I can revert if you would keep this hard couple wiring.

nfalco79:
>Amend an integration test to https://github.com/mojohaus/license-maven-plugin/tree/master/src/it that fails before your fix and passes after your fix.

I can provide an integration test for missingFile but not for missingFileUrl because since HttpRequester does not support URLs other than http/s. This mean that if I provide a file URL it does not work. I should in a prebuild script in groovy create a socket on a free port (to pass in some way to the pom.xml) with a FileServlet that provide the THIRD-PARTY.properties file, or with a maven plugin...it's to much time to spend for just one line fix.

If the implemented integration test just for missingFile (actually not covered by any other ITs) is enough, ok otherwise I have to change HttpRequester to support also file URL protocol (implemented in [this branch](https://github.com/nfalco79/license-maven-plugin/tree/feature/url)) and the IT test will cover both strategies. For 2 missing licenses, one is provided by the missingFile and the other one by missingFileUrl.

Your thought?

nfalco79:
Fix unwanted deletions in IT tests.

nfalco79:
@ppalaga any other request changes? What about generic URL support?

src/it/aggregate-add-third-party-missing-file/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+  <packaging>pom</packaging>
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <license.verbose>true</license.verbose>
+    <license.failOnMissing>true</license.failOnMissing>
+    <license.useMissingFile>true</license.useMissingFile>
+    <!--
+    <license.missingFileUrl>file://${project.basedir}/THIRD-PARTY.properties</license.missingFileUrl>
+     -->
```
ppalaga: Please remove the commented code or explain why it is commented.

src/it/aggregate-add-third-party-missing-file/postbuild.groovy
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+
+file = new File(basedir, 'target/generated-sources/license/THIRD-PARTY.txt');
+assert file.exists();
+content = file.text;
+assert content.contains('(The Apache Software License, Version 2.0) Jettison (org.codehaus.jettison:jettison:1.1 - no url defined)');
+assert content.contains('(The JSON License) JSON (JavaScript Object Notation) (org.json:json:20070829 - http://www.json.org/java/index.html)');
+
+return true;
```
ppalaga: `return true` is redundant after an assert, so please remove `return true`.

src/it/aggregate-add-third-party-missing-file/pom.xml
nfalco79: in my previous post I comment that the test case only cover test missingFile but **not** missingFileUrl (subject of changes) because the plugin only cover HTTP URL. I do not have idea how to provide a valid http server up and running.
So that line it's a reminder that could be de-comment if you agree with the changes on [this branch](https://github.com/nfalco79/license-maven-plugin/tree/feature/url)) where support generic URL. In that branch enable the line above to test both strategies. For 2 missing licenses, one is provided by the missingFile and the other one by missingFileUrl.

src/it/aggregate-add-third-party-missing-file/postbuild.groovy
nfalco79: ok, I will remove. I had copy from other IT test where most of them ends with `return true`

src/it/aggregate-add-third-party-missing-file/pom.xml
nfalco79: PR #249 has that line not commented

ppalaga:
@nfalco79 could you please close this one if #249 is supposed to replace it?

nfalco79:
In the integration test the missingFileUrl point to an not existing host "this.url.does.not.exists"

I expect the mojo fails with a MojoException that say "Unknown host this.url.does.not.exists" (exception message depends from the JVM). Without the fix `missingFileUrl = mojo.missingFileUrl;` the build pass because missingFileUrl is ignored

nfalco79:
@ppalaga I hope the implemented test case is sufficient to demostrate that the missingFileUrl was a missed in the copy from AggregatorAddThirdPartyMojo to AddThirdPartyMojo in initFromMojo method.

Anyway if you have a look to the initFromMojo method it's evident that the goal of that method is copy all attributes from a given mojo to the actual and missingFileUrl was simply forget.

src/it/aggregate-add-third-party-missing-file/invoker.properties
```diff
@@ -0,0 +1,23 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Lesser Public License for more details.
+#
+# You should have received a copy of the GNU General Lesser Public
+# License along with this program.  If not, see
+# <http://www.gnu.org/licenses/lgpl-3.0.html>.
+# #L%
+###
+invoker.goals=clean license:aggregate-add-third-party
+invoker.failureBehavior=fail-never
```
ppalaga: Let's remove this line. We want to see the failure after applying the fix for #245 and we want to see this pass after applying the fix for #134.

nfalco79:
Closed since #245 is not a bug. Maybe it was a misconfiguration in our test to adopt this plugin in our society and in debug we think to find the issue.

Debugging again now we see that the licenseFileUrl is considered just in aggregate mojo after gather missingFile (with a relocated path) in all submodules by the AddThirdPartyMojo.

Sorry to waste your time.

PR #249, opened 2019-02-12
Fix #134 Support generic URL in attributes that ask for URLs

### LINKED ISSUE #134: Can't use any url in missingFileUrl

nfalco79:
Support generic URL for attributes that requires URL:
- licenseMergesUrl
- missingFileUrl
- overrideUrl

The purpose of this PR is support generic URL protocols


nfalco79:
Let me explain in details the integration test. It use both functionality. It cover generic URL feature (#134) and missingFileUrl bug (#245).

There are 2 missing licences:
- jettison
- json

The jettison missing license definition is get by missingFile (all work as expected) without any changes.
The json missing license definition is get by missingFileUrl (subject of all these changes)

What is the point?

* At [ppalaga@4cbc45a](https://github.com/ppalaga/mojohaus-license-maven-plugin/commit/4cbc45aa497106cc541e29b11fbba84a2b740bd9) the test fails, because the missingFileUrl is not take care. If you put any http URL. It will not fails because the attribute is not considered.
* At [pplaga@c1835ab](https://github.com/ppalaga/mojohaus-license-maven-plugin/commit/c1835abbe2c3c61e99336e9f55396ffba9149a49) the test pass because it consider missingFileUrl and it resolve against the file:// protocoll

I can reorder the commit with the fix for (#245) with the missingFileUrl only that point to any foo URL. Since that field is not taken into account it will fails because there are some missing licenses.
With only the fix `missingFileUrl = mojo.missingFileUrl;` the field is now taken into account but should point to a valid http URL that I could not provide. So the changes for (#134) enable me to put a file URL with those licenses that now are taken and make the integration test pass.

To make more coverage I had mix missingFileUrl with also missingFile feature (not covered by any IT tests)

ppalaga:
> * At [ppalaga@4cbc45a](https://github.com/ppalaga/mojohaus-license-maven-plugin/commit/4cbc45aa497106cc541e29b11fbba84a2b740bd9) the test fails, because the missingFileUrl is not take care. If you put any http URL. It will not fails because the attribute is not considered.

As I said, I see the test passing at [ppalaga@4cbc45a](https://github.com/ppalaga/mojohaus-license-maven-plugin/commit/4cbc45aa497106cc541e29b11fbba84a2b740bd9). And from what you say it looks like you agree it should not pass. Not sure what I do differently if you say it fails for you. Let me create PR and let the CI decide who is right.

ppalaga:
> Let me create PR and let the CI decide who is right.

Here https://github.com/mojohaus/license-maven-plugin/pull/256

ppalaga:
BTW, @ptahchiev said in https://github.com/mojohaus/license-maven-plugin/pull/162#issue-234752690

> resolveUnsafeDependenciesFromFile method incorrectly checks if missingLicenses file exists and only then reads missingFileUrl property.

So maybe you should try to remove missingLicenses file from the test to create the situation where the  missingFileUrl is ignored?

ppalaga:
https://github.com/mojohaus/license-maven-plugin/pull/256 passed

nfalco79:
> 
> 
> BTW, @ptahchiev said in [#162 (comment)](https://github.com/mojohaus/license-maven-plugin/pull/162#issue-234752690)
> 
> > resolveUnsafeDependenciesFromFile method incorrectly checks if missingLicenses file exists and only then reads missingFileUrl property.
> 
> So maybe you should try to remove missingLicenses file from the test to create the situation where the missingFileUrl is ignored?

The test in the PR #246 has the missingFileUrl that point to an not existing host "this.url.does.not.exists"

I expect the mojo fails with "Unknown host this.url.does.not.exists" (exception message depends from the JVM). Without the fix `missingFileUrl = mojo.missingFileUrl;` the build pass because missingFileUrl is ignored

nfalco79:
I had separate this PR with the other one just cloning the IT test. This test one is based on add-third-party goal (not affected by #245) instead of aggregate-add-third-party. It try to load a a missing license file by a file URL.

nfalco79:
This PR was separated from the other one so can be review alone.

nfalco79:
I have to add a test case for classpath URL

nfalco79:
I have change the integration test to use classpath protocol to load the missingFileUrl inside a jar in the plugin classpath.

I had to change the initial strategy to register custom URLStreamHandlers because at runtime the URL class when lookup an handler for a given protocol performs a Class.forName. But instead use the maven classloader (ClassRealm) it get the parent (java.class.path) and then system classloader that are not able to load the class in the maven plugin. For this reason any Handler registered through the system property "java.protocol.handler.pkgs" did not work.

src/it/add-third-party-missing-file/pom.xml
```diff
@@ -0,0 +1,80 @@
...
+  <name>License Test :: add-third-party-missing-file</name>
+
+  <packaging>jar</packaging>
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <license.verbose>true</license.verbose>
+    <license.failOnMissing>true</license.failOnMissing>
+    <license.useMissingFile>true</license.useMissingFile>
+    <license.missingFileUrl>classpath:missing-license.properties</license.missingFileUrl>
```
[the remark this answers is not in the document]
nfalco79: ok

src/it/add-third-party-missing-file/prebuild.groovy
```diff
@@ -0,0 +1,45 @@
...
+import java.util.jar.Manifest;
+import java.util.zip.ZipEntry;
+
+JarOutputStream licenseRepo = new JarOutputStream(new FileOutputStream(new File(basedir, "license-repo.jar")));
+
+licenseRepo.putNextEntry(new ZipEntry("missing-license.properties"));
+licenseRepo.write("org.json--json--20070829=The JSON License".getBytes());
+licenseRepo.closeEntry();
+
+licenseRepo.putNextEntry(new ZipEntry("META-INF/MANIFEST.MF"));
```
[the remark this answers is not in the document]
nfalco79: it was to mark from where that jar come

src/license/THIRD-PARTY.properties
```diff
@@ -17,14 +17,14 @@
 # Please fill the missing licenses for dependencies :
 #
 #
-#Tue Aug 22 10:43:57 CEST 2017
```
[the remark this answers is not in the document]
nfalco79: change for mistake, I will do

nfalco79:
> 
> 
> About [006dd52](https://github.com/mojohaus/license-maven-plugin/commit/006dd520f12a381b12bb7f3b036c172787ccef0e):
> 
>     * Please address the comments inline
> 
>     * About `org.codehaus.mojo.license.requester`:
>       
>       * Too many classes for what we need.
>       * Please do your stuff in the current `HttpRequester.` I do not mind if you rename it to e.g. UrlLoader.
>       * Please do not use `URLStreamHandler`. You even do not need to create a URL out of a `classpath:` string.
>       * I'd be OK with supporting just http(s):, file: and classpath:. Those would be implementable with an if statement in the current HttpRequester.
>       * Please add a unit test that checks all schemes (http, https:, file: and classpath:)

The name HttpRequester it's a bit misleading since it handle more request types. What do you think rename HttpRequester as URLRequester?

src/it/add-third-party-missing-file/prebuild.groovy
```diff
@@ -0,0 +1,45 @@
...
+import java.io.File;
+import java.io.FileOutputStream;
+import java.util.jar.JarOutputStream;
+import java.util.jar.Manifest;
+import java.util.zip.ZipEntry;
+
+JarOutputStream licenseRepo = new JarOutputStream(new FileOutputStream(new File(basedir, "license-repo.jar")));
+
+licenseRepo.putNextEntry(new ZipEntry("missing-license.properties"));
+licenseRepo.write("org.json--json--20070829=The JSON License".getBytes());
```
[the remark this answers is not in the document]
ppalaga: > new File(basedir, "license-repo.jar")

I think `basedir + "/target"` would be a better place for the file. Otherwise someone someday is going to add it to git.

ppalaga:
> What do you think rename HttpRequester as URLRequester?

I'd be OK with UrlRequester or UrlLoader.

src/it/add-third-party-missing-file/prebuild.groovy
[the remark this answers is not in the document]
nfalco79: the license-repo.jar file is created already in target/it/add-third-party-missing-file, do you want I move it to target/it/add-third-party-missing-file/target?

nfalco79:
Rebase from master and apply al changes request less one (I did read before last push).

I have done two commit because I found that overrideUrl did not use the HttpRequester.getFromUrl. So to support also that use case I have change a few some method signatures. Methods that returns URL now return String.
The only negative point is there are two interfaces in api package have been changed. That I duplicate those signatures or that I has changed them anyway are not backward compatible.

src/main/java/org/codehaus/mojo/license/utils/UrlRequester.java
```diff
@@ -82,80 +82,85 @@ public static boolean isStringUrl( String data )
...
-                if ( resourceUrl != null )
-                {
-                    result = IOUtils.toString( resourceUrl, utf8 );
-                }
-                else
-                {
-                    throw new MojoExecutionException( "The resource " + resource
-                            + " was not found in the maven plugin classpath" );
-                }
+                resourceUrl = ClassLoader.getSystemResource( resource );
```
ppalaga: The fallback to system classloader is redundant. `UrlRequester.class.getClassLoader()` gives you the plugin classloader whose `getResource()` already delegates to the parent class loader (and thus ultimately to the system classloader). See also https://maven.apache.org/guides/mini/guide-maven-classloading.html

src/main/java/org/codehaus/mojo/license/utils/UrlRequester.java
```diff
@@ -82,80 +82,85 @@ public static boolean isStringUrl( String data )
...
+            }
+            else
+            {
+                throw new IOException( "The resource " + resource
+                        + " was not found in the maven plugin classpath" );
+            }
+        }
+        else if ( "http".equals( protocol ) || "https".equals( protocol ) )
+        {
+            CloseableHttpClient httpClient = HttpClientBuilder.create().build();
```
ppalaga: `CloseableHttpClient` implements `Closeable`. Please use `try (CloseableHttpClient httpClient = HttpClientBuilder.create().build())`

src/main/java/org/codehaus/mojo/license/utils/UrlRequester.java
```diff
@@ -82,80 +82,85 @@ public static boolean isStringUrl( String data )
...
+            HttpGet get = new HttpGet( url );
+            CloseableHttpResponse response = null;
```
ppalaga: Our baseline is Java 7. Plz use `try (CloseableHttpResponse response  = httpClient.execute( get ))` to have less lines.

src/main/java/org/codehaus/mojo/license/utils/UrlRequester.java
```diff
@@ -82,80 +82,85 @@ public static boolean isStringUrl( String data )
...
+            try
+            {
+                response = httpClient.execute( get );
+                result = IOUtils.toString( response.getEntity().getContent(), charset );
```
ppalaga: Using whatever charset there is stored in `charset` might not be the best idea here. You should prefer the charset from `org.apache.http.entity.ContentType.getCharset()` and eventually use your `charset` variable only as a fallback.

src/main/java/org/codehaus/mojo/license/utils/UrlRequester.java
```diff
@@ -82,80 +82,85 @@ public static boolean isStringUrl( String data )
...
+            }
+        }
+        else if ( "http".equals( protocol ) || "https".equals( protocol ) )
+        {
+            CloseableHttpClient httpClient = HttpClientBuilder.create().build();
+            HttpGet get = new HttpGet( url );
+            CloseableHttpResponse response = null;
+            try
+            {
+                response = httpClient.execute( get );
```
ppalaga: You should check the response code and throw an appropriate exception if you get != 200

src/it/add-third-party-missing-file/pom.xml
```diff
@@ -0,0 +1,110 @@
...
+          </dependencies>
+        </plugin>
+      </plugins>
+    </pluginManagement>
+
+    <plugins>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>license-maven-plugin</artifactId>
+        <version>@pom.version@</version>
```
ppalaga: Either remove the `<pluginManagement>` above or remove `<version>` here. One of those is redundant.

src/it/add-third-party-missing-file/pom.xml
```diff
@@ -0,0 +1,110 @@
...
+            </configuration>
+          </execution>
+          <execution>
+            <id>by-file</id>
+            <phase>generate-resources</phase>
+            <goals>
+              <goal>add-third-party</goal>
+            </goals>
+            <configuration>
+              <missingFileUrl>file:///${basedir}/missing-licenses.properties</missingFileUrl>
```
ppalaga: I think both executions write to the same report file. You probably need to configure separate files per execution.

src/it/add-third-party-missing-file/postbuild.groovy
```diff
@@ -0,0 +1,29 @@
...
+ * #L%
+ */
+
+
+file = new File(basedir, 'build.log');
+assert file.exists();
+content = file.text;
+// ensure mojo has been executed two times
+assert content.contains('The JSON License by classpath url');
+assert content.contains('The JSON License by file url');
```
ppalaga: Ah, you check just the log. Plz. check also the report files.

src/main/java/org/codehaus/mojo/license/utils/UrlRequester.java
nfalco79: ok

src/main/java/org/codehaus/mojo/license/utils/UrlRequester.java
nfalco79: You are right, here should be used the encoding in the request/respose headers like Accept-Charset/Content-Type ...
Keep in mind that this code was already in place expect for charset and comment the exception on close connection because could override the main IOException.

src/it/add-third-party-missing-file/postbuild.groovy
nfalco79: ok

src/it/add-third-party-missing-file/pom.xml
nfalco79: ok

src/main/java/org/codehaus/mojo/license/utils/UrlRequester.java
nfalco79: The code was already in place, anyway I changes like you sugget where it takes the payload only if response of the HTTP code if 200 <= code < 300. All 200 codes are valid response (also if empty) not only HTTP_OK.

src/it/add-third-party-missing-file/pom.xml
nfalco79: right

src/it/add-third-party-missing-file/postbuild.groovy
nfalco79: checked only report files since build log in this case does not much sense.

nfalco79:
Changes done. When you are ok I will squash commits

src/main/java/org/codehaus/mojo/license/LicenseMojoUtils.java
```diff
@@ -106,7 +94,7 @@ private static URL prepareUrl( final URL resolvedUrl, final File deprecatedFile,
         {
             try
             {
-                return deprecatedFile.toURI().toURL();
+                return deprecatedFile.toURI().toURL().toString();
```
ppalaga: `deprecatedFile.toURI().toString()`. `.toURL()` is redundant there.

src/main/java/org/codehaus/mojo/license/LicenseMojoUtils.java
```diff
@@ -116,11 +104,11 @@ private static URL prepareUrl( final URL resolvedUrl, final File deprecatedFile,
 
         final Path basedirPath = basedir.toPath();
 
-        if ( url != null && HttpRequester.isStringUrl( url ) )
+        if ( url != null && UrlRequester.isStringUrl( url ) )
         {
             try
             {
-                return new URL( basedirPath.toUri().toURL(), url );
+                return basedirPath.toUri().toURL().toString();
```
ppalaga: `deprecatedFile.toURI().toString()`. `.toURL()` is redundant there.

src/main/java/org/codehaus/mojo/license/LicenseMojoUtils.java
```diff
@@ -134,7 +122,7 @@ private static URL prepareUrl( final URL resolvedUrl, final File deprecatedFile,
         {
             try
             {
-                return defaultPath.toUri().toURL();
+                return defaultPath.toUri().toURL().toString();
```
ppalaga: `deprecatedFile.toURI().toString()`. `.toURL()` is redundant there.

src/test/java/org/codehaus/mojo/license/utils/UrlRequesterTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+
+    @Test
+    public void test_generic_requester() throws Exception
+    {
+        URL res = getClass().getClassLoader().getResource( RESOURCE_NAME );
+        File testFile = fileRule.newFile();
+        FileUtils.copyURLToFile( res, testFile );
+
+        String licenseContent = UrlRequester.getFromUrl( testFile.toURI().toURL().toExternalForm() );
```
ppalaga: `testFile.toURI().toURI().toString()`

src/it/add-third-party-missing-file/pom.xml
ppalaga: Funny that `file:///${basedir}` works. IIRC, ${basedir} tends to be an absolute path, which on Linux gives four slashes - e.g. `file:////path/to/basedir`. I thought, ${project.baseUri} is the only reliable way to test situations like this.

src/test/java/org/codehaus/mojo/license/utils/UrlRequesterTest.java
```diff
@@ -0,0 +1,46 @@
...
+public class UrlRequesterTest
+{
+
+    private static final String RESOURCE_NAME = "org/codehaus/mojo/license/utils/licenses.properties";
+
+    @Rule
+    public TemporaryFolder fileRule = new TemporaryFolder();
+
+    @Test
+    public void test_classpath_requester() throws Exception
```
ppalaga: Plz. use camelCase as usual in Java method names.

src/test/java/org/codehaus/mojo/license/utils/UrlRequesterTest.java
```diff
@@ -0,0 +1,46 @@
...
+    {
+        String licenseContent = UrlRequester.getFromUrl( "classpath:" + RESOURCE_NAME );
+        Assert.assertEquals( "license1=This is mine!", licenseContent.trim() );
+    }
+
+
+    @Test
+    public void test_generic_requester() throws Exception
```
ppalaga: Plz. use camelCase as usual in Java method names.

src/test/java/org/codehaus/mojo/license/utils/UrlRequesterTest.java
```diff
@@ -0,0 +1,46 @@
...
+        URL res = getClass().getClassLoader().getResource( RESOURCE_NAME );
+        File testFile = fileRule.newFile();
+        FileUtils.copyURLToFile( res, testFile );
+
+        String licenseContent = UrlRequester.getFromUrl( testFile.toURI().toURL().toExternalForm() );
+        Assert.assertEquals( "license1=This is mine!", licenseContent.trim() );
+    }
+
+    @Test
+    public void test_registration_of_classpath_handler()
```
ppalaga: Plz. use camelCase as usual in Java method names.

src/test/java/org/codehaus/mojo/license/utils/UrlRequesterTest.java
nfalco79: ok. Usually I prefer that way to write method test because increase readibility

src/main/java/org/codehaus/mojo/license/LicenseMojoUtils.java
nfalco79: For what I can read in the javadoc of `Path.toUri()`  There is no guarantee that the URI may be used to construct a java.io.File that means the I could got an Uri that could not be be handled by standard URL handlers. Moreover the `toUri()` does not automatically escape characters that are illegal in URLs.

This code was already in place and I would not change. With handler approach (maybe think different than I implemented) these changes would not be necessary.

nfalco79:
Changes done. The `.toUri().toURL()` as explained in inline comment are left as they were

src/main/java/org/codehaus/mojo/license/LicenseMojoUtils.java
ppalaga: > in the javadoc of `Path.toUri()` There is no guarantee that the URI may be used to construct a java.io.File

Do we ever directly create a `java.io.File` out of the resulting URI string? I am not aware of any such execution path. 

What we do is using the string returned by `LicenseMojoUtils.prepareUrl()` to construct a URL in `UrlRequester.getFromUrl()` and I think that should work well with `defaultPath.toUri().toString()`. 

The JavaDoc in `Path.toUri()` says that something like `new File(new URI(stringPath))` may not work for UNC stringPaths. But as I said, our path of execution is different - it's something like `new URL(new URL(path.toUri().toString())).openStream()`. 

Actually the best way to figure out would be to write a simple test checking if `path.toUri().toString()` and `path.toUri().toURL().toString()` are equal for UNC paths. If they are, we can safely remove `.toURL()`. Let me try that.

ppalaga: Here is the test: https://github.com/mojohaus/license-maven-plugin/pull/263

ppalaga: The test is passing on Windows and fails on Linux (which is kinda expected because UNC paths are not supposed to work without being mounted). So I think the test proves that `.toURL()` can be removed. WDYT?

ppalaga:
The removal of `.toURL()` is the last open issue from my PoV. And please squash all your commits.

nfalco79:
done

ppalaga:
Thanks, @nfalco79, good work! 

Let's ignore the last occurence of `toURL()` in `UrlRequesterTest` for now ;)

PR #276, opened 2019-02-21
Fix #134 when missingFileUrl contains more lines

### LINKED ISSUE #134: Can't use any url in missingFileUrl

nfalco79:
With latest PR I remove for mistake the `.split( "\n" )` from missingFileUrl than means now read only the first line of the content of URL.

I had update the integration test with one more missing license dependency so the file loaded by URL contains two lines.


ppalaga:
Thanks for your contribution @nfalco79 !

PR #279, opened 2019-02-21
Fix #278 Add timeout parameters to AbstractDownloadLicensesMojo

### LINKED ISSUE #278: Add timeout parameters to AbstractDownloadLicensesMojo

ppalaga:


nfalco79:
Do you want use these property also to configure the HTTP client?

ppalaga:
> Do you want use these property also to configure the HTTP client?

The new params are used to configure the HTTP client used by AbstractDownloadLicensesMojo. Other HTTP clients instantiated elsewhere are out of the scope of the current PR.

PR #280, opened 2019-02-21
Fix #277 aggregate-add-third-party does not use missingFileUrl parameter

### LINKED ISSUE #277: aggregate-add-third-party does not use missingFileUrl parameter
Even though missingFileUrl is set and resolves missing license of a bundle project's dependency, the aggregate-add-third-party goal fails checking for license of that dependency.
I'll show you in an integration test

elballa:
Integration test reproducing the issue

### DIFF
--- a/src/it/aggregate-add-third-party-missing-file/child/pom.xml
+++ b/src/it/aggregate-add-third-party-missing-file/child/pom.xml
@@ -1,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  #%L
+  License Maven Plugin
+  %%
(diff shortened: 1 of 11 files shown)


elballa:
Hello @ppalaga, let me know if fix is as your expectation

elballa:
We created three integration tests as you asked for.
We found a new problem in using missingFile parameter because it was overrided by default value of deprecated aggregateMissingLicensesFile parameter as explained in the issue #284. So we also provided the solution of that issue by removing the assignment of the default value.

ppalaga:
Replaced by https://github.com/mojohaus/license-maven-plugin/pull/286

PR #281, opened 2019-02-21
Fix #271 aggregate-add-third-party evaluate acceptPomPackage also for current project

### LINKED ISSUE #271: acceptPomPackage does not have effect on the parent project when aggregate-add-third-party
Hi, we have a maven project (packaging pom) that aggregate only artifacts to be delivered to a client.
To do this usually we define dependencies on a project of with pom packaging and run the assembly plugin (or depedency-copy-dependency) to gathers all of them.
Now our department would realize a report to double check all licenses to give to a client.
The acceptPomPackage documents "To execute or not this mojo if project packaging is pom." but it is not executed if the project is that where we run the mvn command.

Looking the code the aggregate-add-third-party cycles on each project in the reactor but skip itself. This mean that only child module of kind pom are considered. From the documentation I expect it run a forked add-third-party also on the parent pom (the project from which I run mvn command).

nfalco79:
The aggregate-add-third-party goal now evaluate acceptPomPackage also for the current project like add-third-party does.

### DIFF
--- a/src/it/aggregate-add-third-party-no-children/invoker.properties
+++ b/src/it/aggregate-add-third-party-no-children/invoker.properties
@@ -1,0 +1,23 @@
+###
+# #%L
+# License Maven Plugin
+# %%
+# Copyright (C) 2019, Falco Nikolas
+# %%
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


nfalco79:
Yes I will do thos night

ppalaga:
> Yes I will do thos night

I am doing it now.

ppalaga:
Replaced by https://github.com/mojohaus/license-maven-plugin/pull/287

nfalco79:
I do not see the 1.18 release on maven central. Is it correct? Usually within 24 hours the release is available by oss.sonatype.org

ppalaga:
Voting is in progress https://groups.google.com/d/msg/mojohaus-dev/1AuEwCubmrg/RAajxHR2AQAJ

PR #286, opened 2019-02-23
Fix #277 aggregate-add-third-party does not use missingFileUrl parameter

### LINKED ISSUE #277: aggregate-add-third-party does not use missingFileUrl parameter

ppalaga:
@elballa these are all your changes from https://github.com/mojohaus/license-maven-plugin/pull/280 , I just just reorderd them to make more sense.


ppalaga:
@elballa thanks for the contribution!

PR #290, opened 2019-02-25
Fix #253 Add licenseMergesUrl also in report goals

### LINKED ISSUE #253: Add licenseMergesUrl also in report goals
Existing report goals:
- third-party-report
- aggregate-third-party-report

does not support licenseMergesUrl like their conterpart:
- add-third-party
- aggregate-add-third-party

elballa:
There is still something missing in this fix #253. It should be included in next release because licenseMergeUrl in not reliable if there is more then one raw in content of file linked by licenseMergeUrl

### DIFF
--- a/src/main/java/org/codehaus/mojo/license/AbstractThirdPartyReportMojo.java
+++ b/src/main/java/org/codehaus/mojo/license/AbstractThirdPartyReportMojo.java
@@ -392,7 +392,7 @@ public abstract class AbstractThirdPartyReportMojo extends AbstractMavenReport
             getLog().warn( "" );
             if ( UrlRequester.isStringUrl( licenseMergesUrl ) )
             {
-                licenseMerges = Arrays.asList( UrlRequester.getFromUrl( licenseMergesUrl ) );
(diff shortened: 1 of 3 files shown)


elballa:
There is still something missing in fix #253. It should be included in next release because licenseMergeUrl in not reliable if there is more then one raw in content of file linked by

src/main/java/org/codehaus/mojo/license/AbstractThirdPartyReportMojo.java
```diff
@@ -392,7 +392,7 @@ protected void init()
             getLog().warn( "" );
             if ( UrlRequester.isStringUrl( licenseMergesUrl ) )
             {
-                licenseMerges = Arrays.asList( UrlRequester.getFromUrl( licenseMergesUrl ) );
+                licenseMerges = Arrays.asList( UrlRequester.getFromUrl( licenseMergesUrl ).split( "\n" ) );
```
ppalaga: `[\n\r]+` would perhaps be safer

elballa: done

Issue #201, opened 2019-01-11
add an option to avoid "remove it from the missing file" in the build log


TorstenKruse:
In our multi-module project we use the license-plugin with the missingFileUrl option. 
We created a company-wide missing file which contains GAV and license information for all the dependencies used in our projects for which the license information is not supplied by the dependency itself. So we do not need to gather this information for every single project.

As a result the global missingFile contains more GAV than used in a single project.

This causes the build log to contain a lot of lines with
> [08:29:56] [WARNING] dependency [<GAV>] does not exist in project, remove it from the missing file.

For example for one of the current projects there are 7477 such lines in the build log. 

This makes it really hard to find any other information in the build log.

So I would suggest to add an option to decide if this kind of warning should be logged or not.

Or maybe the warning is completely removed or set to debug, because if does not imply any real problem.

ppalaga:
Looks like a valid use case. Feel free to submit a PR that would contain at least one integration test.

P.S.: Windows CI works now and I am in the process of solving some Windows build/test issues.

digulla:
I tried to fix this using different strategies but I can't find a way to get a `org.codehaus.plexus.logging.Logger` in `AggregatorAddThirdPartyMojo`. Any ideas?

PR #300, opened 2019-02-27
#201 - Reduce log level for unknown dependencies to debug to reduce l…

ercpe:
…og spam when using a shared missing file

https://github.com/mojohaus/license-maven-plugin/issues/201


ppalaga (changes requested):
The proposed solution eliminates the warnings also for users that want to see them and I cannot accept it like this. I propose to solve the issue by adding a parameter much like we did in AbstractDownloadLicensesMojo
https://github.com/mojohaus/license-maven-plugin/blob/1a890f1f1e5911eaee23a4f79455ff482f449822/src/main/java/org/codehaus/mojo/license/AbstractDownloadLicensesMojo.java#L319-L333

I'd recommend to add your own ErrorRemedy-like enum that would initially contain just `warn` (default) and `ignore` so that you not need to implement the other options we have in AbstractDownloadLicensesMojo.

Do not forget to add an integration test.

guidograzioli:
I'm interested in this change, is anybody currently working on it? Happy to help otherwise

src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
```diff
@@ -600,6 +600,20 @@
     @Parameter( property = "project.artifacts", required = true, readonly = true )
     protected Set<Artifact> dependencies;
 
+
+    /**
+     * What to do on defined files not found in project. The possible values are:
+     * <li>
+     *   <ul>{@link UnkownFileRemedy#debug}: unkown files are logged debug</ul>
```
ercpe: Typo throughout the change Unkown -> Unknown

src/main/java/org/codehaus/mojo/license/api/DefaultThirdPartyTool.java
```diff
@@ -552,7 +554,7 @@ public SortedProperties loadUnsafeMapping( LicenseMap licenseMap,
             MavenProject project = artifactCache.get( id );
             if ( project == null )
             {
-                LOG.warn( "dependency [{}] does not exist in project.", id );
+                handleUnkownDependency(unkownFileRemedy,  "dependency [" + id + "] does not exist in project." );
```
ercpe: Probably cleaner to hand `id` down to `handleUnkownDependency` instead of passing the log message around.

wuwu2000:
postbuild integration tests seem to run not 100% stable -,.-

wuwu2000:
can someone tell me why some builds fail randomly?

wuwu2000:
@ppalaga Could you see why this jdk7 build fails please?

PR #304, opened 2019-03-02
Fix #303 Aggregate-add-third-party mojo is not including optional dependencies by default

### LINKED ISSUE #303: Aggregate-add-third-party mojo is not including optional dependencies by default
aggregate-add-third-party mojo is not taking care of the includeOptional parameter because its value is not passed to child modules when child mojo is going to be initialized (initFromMojo method call)

elballa:


ppalaga:
Thanks @elballa !

PR #316, opened 2019-03-30
Fix network proxy configuration

woglinde:
Unfortunately the currently used httpClient works
a little bit diffrently with the proxy settings.

With the old configuration the httpClient
allways gets a defaultRequestConfig without a
proxy set and so always tries a directly connection.


ppalaga:
Thanks a lot for the PR. I'd amend it with a test, e.g. using https://github.com/adamfisk/LittleProxy (unless you very much want to do it yourself?)

woglinde:
> Thanks a lot for the PR. I'd amend it with a test, e.g. using https://github.com/adamfisk/LittleProxy (unless you very much want to do it yourself?)

No problem go ahead.

ppalaga:
Superceded by https://github.com/mojohaus/license-maven-plugin/pull/320

PR #320, opened 2019-04-16
Fix #319 Proxy configuration ignored for download licenses

### LINKED ISSUE #319: Proxy configuration ignored for download licenses
Copying the report by @woglinde from https://github.com/mojohaus/license-maven-plugin/pull/316 : 

> Unfortunately the currently used httpClient works
a little bit diffrently with the proxy settings.
>
> With the old configuration the httpClient
allways gets a defaultRequestConfig without a
proxy set and so always tries a directly connection.

The above PR needs an integration test, e.g. using https://github.com/adamfisk/LittleProxy

ppalaga:


ppalaga:
Thanks again, @woglinde !

Issue #219, opened 2019-01-29
The plugin performs artifact resolution first and then includedScopes.


Tibor17:
We have `includedScopes=compile, runtime`.
Nevertheless the plugin resolves test dependencies and should not.

ppalaga:
Please add steps to reproduce, ideally in from of an integration test under https://github.com/mojohaus/license-maven-plugin/tree/master/src/it

Which mojo(s) does this relate to?

ppalaga:
@Tibor17 do you want to add the missing details? Otherwise, I'd close this as non-reproducible.

Tibor17:
It's in our commercial project.
This is performance issue. Behavior is ok.
It may to do with an algorithm of the Maven Core 2.2.1.
While plugin resolves dependencies, all are downloaded and `includedScopes` is done afterwards which is bad because this takes too long and some dependencies fail to download.

ppalaga:
> It may to do with an algorithm of the Maven Core 2.2.1.

The maven-core dependency has `<scope>provided</scope>` https://github.com/mojohaus/license-maven-plugin/blob/master/pom.xml#L172-L176 so you should get the version delivered by your current maven runtime. 

Or you mean that newer maven-core offers an API to "filter before" download?

Tibor17:
@ppalaga 
I mean newer version - `mavenVersion`.
I tried to upgrade the project to 3.0 but it's not only about `mavenVersion`.
I can do it later, not today.

ppalaga:
@Tibor17, could you please check if the current issue was solved by https://github.com/mojohaus/license-maven-plugin/pull/325 ?

PR #325, opened 2019-04-20
Fix #324 Remove unnecessary manual dependency resolution

### LINKED ISSUE #324: AggregateAddThirdParty (and probably anything else using DependenciesTool) resolves unnecessary artifacts
On a project with the following dependencies:
```
        <!-- Depends on old hbase-annotations, which depends on jdk.tools -->
        <dependency>
            <groupId>org.apache.hive.hcatalog</groupId>
            <artifactId>hive-hcatalog-core</artifactId>
            <version>2.3.4</version>
        </dependency>
        <!-- This has a newer hbase-annotations without jdk.tools. This path to hbase-annotations is shorter, so has precedence -->
        <dependency>
            <groupId>org.apache.storm</groupId>
            <artifactId>child2</artifactId>
            <version>${project.version}</version>
        </dependency>
```
where child2 has the dependency
```
<dependency>
            <groupId>org.apache.hbase</groupId>
            <artifactId>hbase-annotations</artifactId>
            <version>2.1.3</version>
        </dependency>
```

the expected behavior is that the license plugin should resolve only hbase-annotations 2.1.3, and thus not resolve jdk.tools. What actually happens is that both hbase-annotations 2.1.3 and 1.1.1 (incldued by hive-hcatalog-core) are resolved. This causes the build to fail on Java 9+.

I'll put up a PR with a test and fix shortly.

srdo:
Please see https://github.com/mojohaus/license-maven-plugin/issues/324

The fix here is to remove the manual dependency resolution in DependenciesTool. We already configure the mojos to require dependency resolution up to test scope, so all dependencies are already resolved and present in the injected MavenProject(s).

I realize that https://github.com/mojohaus/license-maven-plugin/issues/219 calls for programmatic dependency resolution, but I think that will be easier to implement once the plugin is upgraded to newer APIs. For now, the code in DependenciesTool is unnecessary and causes this bug.

I'll put up a PR for upgrading to Maven 3 APIs soon, which will depend on this PR.

### DIFF
--- a/src/it/ISSUE-324/README.txt
+++ b/src/it/ISSUE-324/README.txt
@@ -1,0 +1,5 @@
+Demonstrates that the aggregateAddThirdParty plugin only resolves the actually used version of a dependency, and not all versions that are present in transitive dependencies.
+
+The project contains a child2 that depends on hbase-annotations:2.1.3. It also contains child1 which depends on hive-hcatalog-core:2.3.4. The Hive dependency also depends on hbase-annotations, but in a much older version, and through a longer dependency chain. The old hbase-annotations artifact depends on jdk.tools, which is not present in Java 9+.
+
+The plugin execution will fail if the plugin attempts to resolve jdk.tools on Java 9+. This test is therefore only meaningful on Java 9+, and will pass spuriously on earlier versions.
--- a/src/it/ISSUE-324/child1/pom.xml
+++ b/src/it/ISSUE-324/child1/pom.xml
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.codehaus.mojo.license.test</groupId>
+    <artifactId>test-ISSUE-324</artifactId>
+    <version>@pom.version@</version>
+  </parent>
+  <artifactId>test-ISSUE-324-child1</artifactId>
+  <packaging>jar</packaging>
+
+  <dependencies>
+        <!-- Depends on old hbase-annotations, which depends on jdk.tools -->
+    <dependency>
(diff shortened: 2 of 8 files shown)


srdo:
One of the builds on Appveyor failed, but it's not the test I added, and it doesn't fail for me locally or on Travis. The log on Appveyor just refers to the integration test build log. Do you have access to that log @ppalaga? I don't seem to be able to find it.

ppalaga:
> One of the builds on Appveyor failed, but it's not the test I added, and it doesn't fail for me locally or on Travis. 

Appveyor is Windows. Do you use Windows locally?

> The log on Appveyor just refers to the integration test build log. Do you have access to that log @ppalaga?

No, Appveyor, much like TravisCI has no notion of persistent workspace. The only way to debug this is to add `type path\to\the\log` to `on_failure` of `appveyor.yml`.

I pressed "rebuild" on the failed task, FWIW.

srdo:
I use Windows, but through the Linux subsystem. It looks like the test passes now, maybe it is flaky on WIndows. I'll try running it a couple of times outside the Linux subsystem.

Regarding Appveyor, maybe we could store the target/it directory as artifacts on failure? https://help.appveyor.com/discussions/suggestions/828-store-some-artifacts-in-case-of-failure

srdo:
Ran the tests 5 times locally on Windows with no failures. Maybe it was a network issue while downloading a license?

ppalaga:
> Maybe it was a network issue while downloading a license?

Yeah, that's my theory. Let me review this PR.

> maybe we could store the target/it directory as artifacts on failure? https://help.appveyor.com/discussions/suggestions/828-store-some-artifacts-in-case-of-failure

A PR is welcome.

ppalaga:
Does this commit have any potential to break the backwards compatibility?

ppalaga:
Found no issue with WildFly Camel. This PR will be good to merge once you please reword to `Fix #324 Remove unnecessary manual dependency resolution`

ppalaga:
My quick tests show this as a substantial perf gain a project as large as WildFly Camel. I am getting build time 2:08 using this PR vs. 3:15 using version 1.20.

srdo:
Thanks for reviewing.

Reworded commit message.

I wouldn't expect backward compatibility to be broken. The deleted code wasn't doing anything useful, I think, and any difference to the dependencies resolved by Maven was probably incorrect.

I'll take a look at the appveyor thing soon.

ppalaga:
Thanks @srdo !

PR #327, opened 2019-04-20
Fix #326 Migrate to using Maven 3 APIs and Aether for dependency resolution

### LINKED ISSUE #326: Upgrade to Maven 3 APIs and Aether
It would be good to upgrade to Maven 3 APIs and Aether. It is difficult to implement some features (programmatic dependency resolution for instance) correctly with the old APIs, and they are also deprecated in newer Maven versions.

The plugin could be upgraded to use Maven 3 dependencies, and maven-resolver (Aether).

srdo:
Based on https://github.com/mojohaus/license-maven-plugin/issues/324, please look at that one first.

This migrates the plugin to Maven 3 dependencies, and also replaces the old ArtifactResolver with maven-resolver.

The version requirement here is at least 3.5.0, as maven-resolver-provider is not present in prior versions. For now I've set the plugin to need 3.6.1, as that was what I tested with.

### DIFF
--- a/src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
+++ b/src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
@@ -27,8 +27,6 @@ import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.artifact.repository.ArtifactRepository;
-import org.apache.maven.artifact.resolver.ArtifactNotFoundException;
-import org.apache.maven.artifact.resolver.ArtifactResolutionException;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.MojoFailureException;
 import org.apache.maven.plugin.logging.Log;
@@ -59,6 +57,8 @@ import java.util.Set;
 import java.util.SortedMap;
 import java.util.SortedSet;
 import org.codehaus.mojo.license.api.DependenciesToolException;
+import org.eclipse.aether.resolution.ArtifactResolutionException;
+import org.eclipse.aether.transfer.ArtifactNotFoundException;
 
 /**
  * Abstract mojo for all third-party mojos.
@@ -593,14 +593,6 @@ public abstract class AbstractAddThirdPartyMojo
     protected ArtifactRepository localRepository;
 
     /**
-     * Remote repositories used for the project.
-     *
-     * @since 1.0.0
-     */
-    @Parameter( property = "project.remoteArtifactRepositories", required = true, readonly = true )
-    protected List<ArtifactRepository> remoteRepositories;
-
-    /**
      * The set of dependencies for the current project, used to locate license databases.
      */
     @Parameter( property = "project.artifacts", required = true, readonly = true )
@@ -881,7 +873,8 @@ public abstract class AbstractAddThirdPartyMojo
         if ( helper == null )
         {
             helper = new DefaultThirdPartyHelper( getProject(), getEncoding(), isVerbose(), dependenciesTool,
-                    thirdPartyTool, localRepository, remoteRepositories, getLog() );
+                    thirdPartyTool, getProject().getRemoteArtifactRepositories(),
+                    getProject().getRemoteProjectRepositories(), getLog() );
         }
         return helper;
     }
@@ -890,7 +883,7 @@ public abstract class AbstractAddThirdPartyMojo
       throws ArtifactNotFoundException, IOException, ArtifactResolutionException, MojoExecutionException
     {
         File missingLicensesFromArtifact = thirdPartyTool.resolveMissingLicensesDescriptor( groupId, artifactId,
-                version, localRepository, remoteRepositories );
+                version, getProject().getRemoteProjectRepositories() );
         resolveUnsafeDependenciesFromFile( missingLicensesFromArtifact );
     }
 
--- a/src/main/java/org/codehaus/mojo/license/AbstractLicensesXmlMojo.java
+++ b/src/main/java/org/codehaus/mojo/license/AbstractLicensesXmlMojo.java
@@ -84,14 +84,6 @@ public abstract class AbstractLicensesXmlMojo
     private String licensesOutputFileEncoding;
 
     /**
-     * Location of the local repository.
-     *
-     * @since 1.0
-     */
-    @Parameter( defaultValue = "${localRepository}", readonly = true )
-    protected ArtifactRepository localRepository;
-
-    /**
      * List of Remote Repositories used by the resolver
      *
      * @since 1.0
@@ -108,12 +100,12 @@ public abstract class AbstractLicensesXmlMojo
     protected MavenProject project;
 
     /**
-     * Dependencies tool.
+     * Licensed artifact resolver.
      *
      * @since 1.0
      */
     @Component
-    protected LicensedArtifactResolver dependenciesTool;
+    protected LicensedArtifactResolver licensedArtifactResolver;
 
     private Charset charset;
 
--- a/src/main/java/org/codehaus/mojo/license/AbstractThirdPartyReportMojo.java
+++ b/src/main/java/org/codehaus/mojo/license/AbstractThirdPartyReportMojo.java
@@ -23,7 +23,6 @@ package org.codehaus.mojo.license;
  */
 
 import org.apache.commons.collections.CollectionUtils;
-import org.apache.maven.artifact.repository.ArtifactRepository;
 import org.apache.maven.artifact.resolver.ArtifactNotFoundException;
 import org.apache.maven.artifact.resolver.ArtifactResolutionException;
 import org.apache.maven.doxia.siterenderer.Renderer;
@@ -302,14 +301,6 @@ public abstract class AbstractThirdPartyReportMojo extends AbstractMavenReport
     private String encoding;
 
     /**
-     * Local Repository.
-     *
-     * @since 1.1
-     */
-    @Parameter( property = "localRepository", required = true, readonly = true )
-    private ArtifactRepository localRepository;
-
-    /**
      * The Maven Project.
      *
      * @since 1.1
@@ -434,10 +425,6 @@ public abstract class AbstractThirdPartyReportMojo extends AbstractMavenReport
             throw new MavenReportException( e.getMessage(), e );
         }
         catch ( ProjectBuildingException e )
-        {
-            throw new MavenReportException( e.getMessage(), e );
-        }
-        catch ( InvalidDependencyVersionException e )
         {
             throw new MavenReportException( e.getMessage(), e );
         }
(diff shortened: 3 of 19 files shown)


srdo:
Set required Maven version to 3.5.4, as that is the version used by CI.

ppalaga:
> Set required Maven version to 3.5.4, as that is the version used by CI.

+1

src/main/java/org/codehaus/mojo/license/api/DefaultThirdPartyTool.java
```diff
@@ -754,59 +750,80 @@ private void loadOneGlobalSet( SortedSet<MavenProject> unsafeDependencies, Licen
...
+                DESCRIPTOR_CLASSIFIER, remoteRepositories );
     }
 
     private File resolveArtifact( String groupId, String artifactId, String version,
-            String type, String classifier, ArtifactRepository localRepository, List<ArtifactRepository> repositories )
-                    throws ArtifactResolutionException, IOException, ArtifactNotFoundException
+            String type, String classifier, List<RemoteRepository> remoteRepositories )
+                    throws ArtifactResolutionException
     {
         // TODO: this is a bit crude - proper type, or proper handling as metadata rather than an artifact in 2.1?
```
ppalaga: Shouldn't this comment be removed? It does not seem to make any sense anymore.

src/main/java/org/codehaus/mojo/license/api/DefaultThirdPartyTool.java
```diff
@@ -837,4 +854,10 @@ private File resolveArtifact( String groupId, String artifactId, String version,
         }
         return migrateKeys;
     }
+
+    @Override
+    public void setAetherRepoSession( RepositorySystemSession aetherRepoSession )
+    {
+        this.aetherRepoSession = aetherRepoSession;
```
ppalaga: Making aetherRepoSession mutable in this way does not feel good. If there is no way to persuade plexus to inject it, I vote for removing the aetherRepoSession field and pass it via method parameter.

src/main/java/org/codehaus/mojo/license/api/DependenciesTool.java
```diff
@@ -163,8 +171,25 @@
...
-                    depMavenProject =
-                        mavenProjectBuilder.buildFromRepository( artifact, remoteRepositories, localRepository, true );
+                    /*
+                     * Some POMs may refer to e.g. java.home or java.version,
+                     * so we need to ensure common system properties are set when parsing POMs.
+                     * Using the current system properties seems like a decent solution.
+                     */
+                    Properties props = new Properties();
+                    props.putAll( System.getProperties() );
+                    ProjectBuildingRequest req = new DefaultProjectBuildingRequest()
```
ppalaga: Do we need a new instance of ProjectBuildingRequest for each iteration of the enclosing for loop?

src/main/java/org/codehaus/mojo/license/api/DependenciesTool.java
```diff
@@ -163,8 +171,25 @@
 
                 try
                 {
-                    depMavenProject =
-                        mavenProjectBuilder.buildFromRepository( artifact, remoteRepositories, localRepository, true );
+                    /*
+                     * Some POMs may refer to e.g. java.home or java.version,
+                     * so we need to ensure common system properties are set when parsing POMs.
+                     * Using the current system properties seems like a decent solution.
+                     */
```
ppalaga: I wonder if it wasn't safer to derive the ProjectBuildingRequest from the currently running request. Something like https://github.com/blackducksoftware/hub-maven-plugin/blob/master/src/main/java/com/blackducksoftware/integration/maven/MavenDependencyExtractor.java#L73

src/main/java/org/codehaus/mojo/license/api/DependenciesTool.java
```diff
@@ -239,4 +264,9 @@
         return result;
     }
     // CHECKSTYLE_ON: MethodLength
+
+    public void setAetherRepoSession( RepositorySystemSession aetherRepoSession )
+    {
+        this.aetherRepoSession = aetherRepoSession;
```
ppalaga: Making aetherRepoSession mutable in this way does not feel good. If there is no way to persuade plexus to inject it, I vote for removing the aetherRepoSession field and pass it via method parameter.

src/main/java/org/codehaus/mojo/license/download/LicensedArtifactResolver.java
```diff
@@ -216,4 +231,8 @@ public void loadProjectDependencies( ResolvedProjectDependencies artifacts,
     }
     // CHECKSTYLE_ON: MethodLength
 
+    public void setAetherRepoSession( RepositorySystemSession aetherRepoSession )
+    {
+        this.aetherRepoSession = aetherRepoSession;
```
ppalaga: Making aetherRepoSession mutable in this way does not feel good. If there is no way to persuade plexus to inject it, I vote for removing the aetherRepoSession field and pass it via method parameter.

src/main/java/org/codehaus/mojo/license/api/DefaultThirdPartyTool.java
srdo: Maybe? I understood the comment to say that handling the third party descriptor as an artifact was crude, but I'm not sure what the intended alternative was. I'll remove the comment.

src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
```diff
@@ -872,6 +871,17 @@ public void setExcludedLicenses( String excludedLicenses ) throws MojoExecutionE
         this.excludedLicenses = new ExcludedLicenses( excludedLicenses );
     }
 
+    /**
+     * Set repo session in the components that need it.
+     * @param session The repository system session
+     */
+    public void setAetherRepoSession( RepositorySystemSession session )
```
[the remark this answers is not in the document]
srdo: Maven calls it when injecting the RepositorySystemSession. The idea was to have this setter initialize the fields in the other classes. I couldn't find a way to get the DI framework to inject it.

src/main/java/org/codehaus/mojo/license/api/DefaultThirdPartyTool.java
srdo: I agree, it is not nice. I couldn't find a way to inject it. I'll move it to method parameters.

src/main/java/org/codehaus/mojo/license/api/DependenciesTool.java
srdo: The parameters don't change, so probably not. I'll change it to reuse the request.

src/main/java/org/codehaus/mojo/license/api/DependenciesTool.java
srdo: That looks better, wasn't aware of MavenSession. Will try it out.

src/main/java/org/codehaus/mojo/license/api/DependenciesTool.java
srdo: I think it is better here. Constructing the request shouldn't take any real time, and moving it outside the for loop puts a lot of distance between the declaration and use of the request.

srdo:
Addressed comments, and reworded commit message.

The MavenSession has a getter for the RepositorySystemSession, and can be injected with `@Requirement`, so I just did that rather than add RepositorySystemSession as a parameter everywhere.

src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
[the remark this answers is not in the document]
srdo: Replaced with MavenSession injection

src/main/java/org/codehaus/mojo/license/AbstractLicensesXmlMojo.java
```diff
@@ -184,4 +183,14 @@ protected void writeLicenseSummary( List<ProjectLicenseInfo> deps, File licenses
                                                   writeVersions );
     }
 
+    /**
+     * Set repo session in the components that need it.
+     * @param session The repository system session
+     */
+    public void setAetherRepoSession( RepositorySystemSession session )
```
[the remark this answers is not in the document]
srdo: Replaced with MavenSession injection

src/main/java/org/codehaus/mojo/license/AbstractThirdPartyReportMojo.java
```diff
@@ -650,5 +644,15 @@ public String getEncoding()
         return encoding;
     }
 
+    /**
+     * Set repo session in the components that need it.
+     * @param session The repository system session
+     */
+    public void setAetherRepoSession( RepositorySystemSession session )
```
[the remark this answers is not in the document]
srdo: Replaced with MavenSession injection

src/main/java/org/codehaus/mojo/license/api/DefaultThirdPartyTool.java
srdo: Removed the comment

src/main/java/org/codehaus/mojo/license/api/DefaultThirdPartyTool.java
srdo: Replaced with MavenSession injection

src/main/java/org/codehaus/mojo/license/api/DependenciesTool.java
srdo: Using MavenSession

src/main/java/org/codehaus/mojo/license/api/DependenciesTool.java
srdo: Replaced with MavenSession injection

src/main/java/org/codehaus/mojo/license/download/LicensedArtifactResolver.java
srdo: Replaced with MavenSession injection

srdo:
The test failure is a timeout.

```
[WARNING] Unable to retrieve license from URL 'http://www.opensource.org/licenses/bsd-license.php' for dependency 'org.hamcrest:hamcrest-core:1.3': Read timed out
```

src/main/java/org/codehaus/mojo/license/api/DependenciesTool.java
ppalaga: > Constructing the request shouldn't take any real time

It very much depends on how many iterations the loop has and also on how many sys props are set.

> puts a lot of distance

Sounds like a readability/aesthetics concern? - those are important, but here, I'd favor performance, unless proven that pulling the req out of the loop is technically impossible.

srdo: Sure, but the system properties aren't being copied anymore, since the session change.

It's just a readability concern.  

I'll move it out of the loop.

srdo: Done

ppalaga:
Thanks, looks very good.
Sorry for the last nit pick: Could you please pull the buildReq out of the loop also in LicensedArtifactResolver.loadProjectDependencies()?

srdo:
Yes, done.

ppalaga:
Ouch, an editorconfig violation. Otherwise looks mergeable. Thanks!

ppalaga:
Great work, thanks a lot, @srdo !

PR #333, opened 2019-05-10
Fix #331 overrideUrl doesn't work

### LINKED ISSUE #331: overrideUrl doesn't work
There is a bug in the code which processes the overrideUrl in org.codehaus.mojo.license.LicenseMojoUtils.prepareUrl(String, File, String, File, String):

        final Path basedirPath = basedir.toPath();

        if ( url != null && UrlRequester.isStringUrl( url ) )
        {
            return basedirPath.toUri().toString();
        }

        final Path defaultPath = basedirPath.resolve( defaultFilePath );

should be

        if ( url != null && UrlRequester.isStringUrl( url ) )
        {
            return url;
        }

        final Path basedirPath = basedir.toPath();
        final Path defaultPath = basedirPath.resolve( defaultFilePath );

digulla:
In addition to just checking the result, I've also added logging so users of the plugin can find out what it's loading and from where and why without having to start a debugger.

Code coverage is at 83.9%

Sore point is the DEFAULT_OVERRIDE_THIRD_PARTY which is used in the default path of the code; I could wrap the constant in a getter to switch between "exists" and "missing" cases.

### DIFF
--- a/src/test/java/org/codehaus/mojo/license/LicenseMojoUtilsTest.java
+++ b/src/test/java/org/codehaus/mojo/license/LicenseMojoUtilsTest.java
@@ -1,0 +1,165 @@
+package org.codehaus.mojo.license;
+
+import static org.junit.Assert.*;
+
+import java.io.File;
+
+import org.junit.Test;
+
+public class LicenseMojoUtilsTest
+{
+    private String resolvedUrl;
+    private File deprecatedFile;
+    private String url;
+    private File basedir = new File( "" );
+    private MockLogger log = new MockLogger();
+
+    @Test
+    public void testIsValidNull()
+    {
+        assertFalse( LicenseMojoUtils.isValid( null ) );
+    }
+
+    @Test
+    public void testIsValidEmpty()
+    {
+        // This might be wrong; feel free to change the test when it starts to fail
+        assertTrue( LicenseMojoUtils.isValid( "" ) );
+    }
+
+    @Test
+    public void testIsValidBlank()
+    {
+        // This might be wrong; feel free to change the test when it starts to fail
+        assertTrue( LicenseMojoUtils.isValid( "   " ) );
+    }
+
+    @Test
+    public void testIsValidNonexistingClasspathResource()
+    {
+        assertTrue( LicenseMojoUtils.isValid( "classpath:noSuchResource" ) );
+    }
+
+    @Test
+    public void testIsValidClasspathResource()
+    {
+        assertTrue( LicenseMojoUtils.isValid( "classpath:log4j.properties" ) );
+    }
+
+    @Test
+    public void testIsValidHttpResource()
+    {
+        assertTrue( LicenseMojoUtils.isValid( "http://foo/bar/baz" ) );
+    }
+
+    @Test
+    public void testPrepareThirdPartyOverrideUrlNull()
+    {
+        String actual = LicenseMojoUtils.prepareThirdPartyOverrideUrl( resolvedUrl, deprecatedFile, url, basedir, log );
+        assertEquals( LicenseMojoUtils.NO_URL, actual );
+    }
+
+    @Test
+    public void testPrepareThirdPartyOverrideUrlBothOverrides()
+    {
+        deprecatedFile = new File( "src/test/resources/overrides.properties" );
+        url = "classpath:overrides.properties";
+        try
+        {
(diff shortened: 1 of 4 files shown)


src/test/java/org/codehaus/mojo/license/LicenseMojoUtilsTest.java
```diff
@@ -0,0 +1,162 @@
...
+    @Test
+    public void testIsValidNull()
+    {
+        assertFalse( LicenseMojoUtils.isValid( null ) );
+    }
+
+    @Test
+    public void testIsValidEmpty()
+    {
+        // TODO is this correct?
```
ppalaga: Yes, let's assume it is correct unless proven otherwise. Please remove the TODO.

src/test/java/org/codehaus/mojo/license/LicenseMojoUtilsTest.java
```diff
@@ -0,0 +1,162 @@
...
+        assertTrue( LicenseMojoUtils.isValid( "" ) );
+    }
+
+    @Test
+    public void testIsValidBlank()
+    {
+        // TODO is this correct?
```
ppalaga: Yes, let's assume it is correct unless proven otherwise. Please remove the TODO.

src/test/java/org/codehaus/mojo/license/LicenseMojoUtilsTest.java
```diff
@@ -0,0 +1,162 @@
...
+    @Test
+    public void testIsValidHttpResource()
+    {
+        assertTrue( LicenseMojoUtils.isValid( "http://foo/bar/baz" ) );
+    }
+
+    @Test
+    public void testPrepareThirdPartyOverrideUrlNull()
+    {
+        LicenseMojoUtils.prepareThirdPartyOverrideUrl( resolvedUrl, deprecatedFile, url, basedir, log );
```
ppalaga: Should there be no assertion about the value returned by `prepareThirdPartyOverrideUrl()`?

src/test/java/org/codehaus/mojo/license/LicenseMojoUtilsTest.java
```diff
@@ -0,0 +1,162 @@
...
+    
+    @Test
+    public void testPrepareThirdPartyOverrideUrlBothOverrides()
+    {
+        deprecatedFile = new File( "src/test/resources/overrides.properties" );
+        url = "classpath:overrides.properties";
+        try
+        {
+            LicenseMojoUtils.prepareThirdPartyOverrideUrl( resolvedUrl, deprecatedFile, url, basedir, log );
+        }
```
ppalaga: If the IllegalArgumentException is expected, the test should fail in case the exception is not thrown. So you probably want to add something like `Assert.fail(IllegalArgumentException.class.getSimpleName() + " expected")` on line 69

src/test/java/org/codehaus/mojo/license/LicenseMojoUtilsTest.java
digulla: I've added a comment: `This might be wrong; feel free to change the test when it starts to fail` so the next person will not hesitate to fix the test.

src/test/java/org/codehaus/mojo/license/LicenseMojoUtilsTest.java
digulla: See above

src/test/java/org/codehaus/mojo/license/LicenseMojoUtilsTest.java
digulla: Fixed

src/test/java/org/codehaus/mojo/license/LicenseMojoUtilsTest.java
digulla: Done.

digulla:
Just to make sure: When you ask me to "squash the commit", I do that locally and then force-push, right?

ppalaga:
> Just to make sure: When you ask me to "squash the commit", I do that locally and then force-push, right?

Yes.

digulla:
I'm wondering why this issue is still in the state "Changes requested". Did I miss something?

ppalaga:
Thanks @digulla !

PR #335, opened 2019-05-10
Feature/201 omit unknown dependencies

digulla:
First attempt to make logging of unknown dependencies configurable.
This is way too much code but I had to introduce new objects or checkstyle would fail the build (too many parameters per method).
Also, I couldn't use plexus loggers and had to switch to slf4j instead because in the Mojo, I can't get a plexus logger. And in other parts, I can only call getLog() which is another beast. Logging is a mess.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -301,10 +301,9 @@
     </dependency>
 
     <dependency>
-      <groupId>log4j</groupId>
-      <artifactId>log4j</artifactId>
-      <version>1.2.17</version>
-      <scope>runtime</scope>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+      <version>1.7.26</version>
     </dependency>
 
     <dependency>
--- a/src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
+++ b/src/main/java/org/codehaus/mojo/license/AbstractAddThirdPartyMojo.java
@@ -39,6 +39,7 @@ import org.codehaus.mojo.license.api.DependenciesTool;
 import org.codehaus.mojo.license.api.ThirdPartyHelper;
 import org.codehaus.mojo.license.api.ThirdPartyTool;
 import org.codehaus.mojo.license.api.ThirdPartyToolException;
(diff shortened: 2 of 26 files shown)


ppalaga:
Yes, it is too much code. Please feel free to use `CHECKSTYLE_OFF: ParameterNumber` and `CHECKSTYLE_ON: ParameterNumber` as we already do elsewhere: https://github.com/mojohaus/license-maven-plugin/blob/master/src/main/java/org/codehaus/mojo/license/api/ArtifactFilters.java#L55-L61

and pass the Logger as an argument.

What problem is `SortedPropertiesProvider` and its impls solving?

digulla:
The `SortedPropertiesProvider` allows the Mojo to decide where to load stuff from. In the PR, that's the missing licenses properties file. It can be a file or a URL or nothing. That way, the Mojo creates the correct provider depending on the config and passes it to the rest of the code which just asks for the `SortedProperties` without having to pass missingFile and missingFileUrl through many methods.

I think the same approach could be used in other placed (overrides) but I ran out of time to change it there, too.

How should we proceed?

I would really like to clean up the code so that the Mojo creates helper objects instead of passing all the config options around.

Also, I don't like the logging as it is right now. In the Mojo, getLog() is being used. In other places, a Plexus logger is passed around. It feels wrong to pass a logger as parameter; the code in Maven is only doing that because the Mojo / Plexus API was forcing you to do this. With Maven 3.1, slf4j is available which allows every class to ask for a logger. So there is no need to pass it via parameter. This should also work in the Mojo and components.

Time permitting, I could try to create a patch which unifies all logging to use slf4j. But that will be a bigger refactoring and I'm not going to spend time on this unless you approve.

ppalaga:
The objective of this PR should be to introduce one mojo parameter and based on that, skip some logger calls. 1060 new lines of code is not proportional to that task. I am open to discuss refactoring but please open separate issues for that and let's discuss there first. I am against mixing major refactoring with a small feature in a single PR. Having said that, I am sorry, I cannot accept this PR in its present form. Please simplify it and concentrate on the primary objective.

digulla:
I don't see a way to implement this without some code cleanup, first.

PR #336, opened 2019-06-04
Update the CDDLv1 license header

rez-:
The CDDL v1.0 license header contained a wrong URL and several typos.

[edit]
We could also use the SPDX URL: https://spdx.org/licenses/CDDL-1.0.html

### DIFF
--- a/src/main/resources/META-INF/licenses/cddl_v1/header.txt
+++ b/src/main/resources/META-INF/licenses/cddl_v1/header.txt
@@ -1,15 +1,9 @@
-COMMON DEVELOPMENT AND DISTRIBUTION LICENSE (CDDL) Version 1.0
+The contents of this file are subject to the terms of the
+Common Development and Distribution License (the "License").
+You may not use this file except in compliance with the License.
 
-This program is free software: you can redistribute it and/or modify
-it under the terms of the Common Development and Distrubtion License as
-published by the Sun Microsystems, either version 1.0 of the
-License, or (at your option) any later version.
+You can obtain a copy of the License at
+https://opensource.org/licenses/CDDL-1.0.
+See the License for the specific language governing permissions
+and limitations under the License.
 
-This program is distributed in the hope that it will be useful,
-but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


ppalaga:
@rez- I would not mind fixing "a wrong URL and several typos" but you actually provide a fully new text. Could you please justify the appropriateness of the change somehow?

rez-:
You're right.
I've been searching around a lot, and found these references to the license header of CDDL1:
- https://bugster.forgerock.org/jira/browse/OPENDJ-1222
- https://zinascii.com/pub/illumos/gate/6205/usr/src/tools/scripts/checkpaths.1onbld.html
- https://github.com/Evolveum/openicf/blob/master/framework/java/connector-framework/src/main- /java/org/identityconnectors/common/script/ScriptExecutor.java
- https://github.com/OpenIdentityPlatform/OpenDJ/blob/290ce62933ac63254c98140d609244a59230cec1/opendj-ldap-sdk/src/main/java/com/forgerock/opendj/ldap/ASN1BufferReader.java
- https://github.com/CedricGatay/portlet-container-custom/blob/master/context/src/java/com/sun/portal/portletcontainer/context/registry/PortletRegistryContextFactory.java

What it comes down to is this text:
`/*
 * CDDL HEADER START
 *
 * The contents of this file are subject to the terms of the
 * Common Development and Distribution License, Version 1.0 only
 * (the "License").  You may not use this file except in compliance
 * with the License.
 *
 * You can obtain a copy of the license at legal-notices/CDDLv1_0.txt
 * or http://forgerock.org/license/CDDLv1.0.html.
 * See the License for the specific language governing permissions
 * and limitations under the License.
 *
 * When distributing Covered Code, include this CDDL HEADER in each
 * file and include the License file at legal-notices/CDDLv1_0.txt.
 * If applicable, add the following below this CDDL HEADER, with the
 * fields enclosed by brackets "[]" replaced with your own identifying
 * information:
 *      Portions Copyright [yyyy] [name of copyright owner]
 *
 * CDDL HEADER END
 */`

So my update is still incorrect. Perhaps we should also remove the file reference and update the link to SPDX?

tchemit:
@rez- I just revert it, need more input to make the change.

PR #344, opened 2019-07-03
Fix #228 Add .gitattributes

### LINKED ISSUE #228: Project doesn't build out of the box on Windows
This is to discuss the need of a `.gitattributes` file.

`git config --list --show-origin` shows that `core.autocrlf` is not defined. I think the default is `true`.

How about dropping `.editorconfig` and configuring Git in such a way that it will work for everyone out of the box, no matter how `autocrlf` is configured? I have about 50 Git projects from all kinds of sources checked out right now. Your project is the only one which doesn't build after checkout because of EOL issues.

kingle:
This worked for me to fix the build on Windows.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+*        text eol=lf
+*.cmd    text eol=crlf


ppalaga:
Very good @kingle , I wanted to see something as minimal as this. Let me check if it works on Linux too.

kingle:
Note: for someone on Windows that pulls this update from an existing checkout -- the existence of this file doesn't automatically change anything in their current working directory.  They will need to re-clone or `git reset --hard` for this to take.

ppalaga:
What are the symptoms for needing to re-clone or `git reset --hard`? I wonder if we should put a note on that to the README?

ppalaga:
Thanks, @kingle !

PR #348, opened 2019-08-05
Fix #139 - "Can't find license-maven-plugin"

### LINKED ISSUE #139: "Can't find license-maven-plugin" in 1.16 if plugin isn't configured in the POM
If you try to run `mvn license:aggregate-add-third-party` without adding the plugin to the root / parent POM first, you get the error `Can't find license-maven-plugin` throws in this line:

https://github.com/mojohaus/license-maven-plugin/blob/70b96dfd6db8b5bde085567ea3cb2ef1bdef2d83/src/main/java/org/codehaus/mojo/license/AggregatorAddThirdPartyMojo.java#L197

I see two solutions:

1. Enhance the error message: "Please add the plugin to your POM"
2. Put the three values in a property file (using resource filtering) and load that file from the classpath.

AndyGee:
Use the PluginDescriptor.
Most painful checkstyle.

### DIFF
--- a/src/it/add-third-party-no-encoding/invoker.properties
+++ b/src/it/add-third-party-no-encoding/invoker.properties
@@ -19,5 +19,5 @@
 # <http://www.gnu.org/licenses/lgpl-3.0.html>.
 # #L%
 ###
-invoker.goals=clean license:add-third-party
+invoker.goals=clean ${project.groupId}:${project.artifactId}:${project.version}:add-third-party
 invoker.failureBehavior=fail-fast
--- a/src/main/java/org/codehaus/mojo/license/AggregatorAddThirdPartyMojo.java
+++ b/src/main/java/org/codehaus/mojo/license/AggregatorAddThirdPartyMojo.java
@@ -33,6 +33,7 @@ import org.apache.commons.collections.CollectionUtils;
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.model.Plugin;
 import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugin.descriptor.PluginDescriptor;
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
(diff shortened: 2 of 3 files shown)


src/main/java/org/codehaus/mojo/license/AggregatorAddThirdPartyMojo.java
```diff
@@ -203,11 +204,30 @@ protected void doAction()
         }
         if ( groupId == null )
         {
-            throw new IllegalStateException( "Can't find license-maven-plugin" );
+            try
+            {
+                final PluginDescriptor pd = ( PluginDescriptor ) getPluginContext().get( "pluginDescriptor" );
+                groupId = pd.getGroupId();
+                artifactId = pd.getArtifactId();
+                version = pd.getVersion();
```
ppalaga: Isn't this actually the preferred way to get the `g:a:v` of the current plugin? If so, I guess we can remove the attempts above (startig at https://github.com/mojohaus/license-maven-plugin/pull/348/files#diff-38b03aa7a6c189ea0f7e814313a14682R182 ) to get it via ` project.getPluginArtifactMap()` and `project.getPluginManagement().getPluginsAsMap().get( "org.codehaus.mojo:license-maven-plugin" )` ?

ppalaga:
OK, now I see there is 0374cf4 that delivers the test coverage.

src/main/java/org/codehaus/mojo/license/AggregatorAddThirdPartyMojo.java
AndyGee: On premise, never touch a running system. I just didn't want to throw that away. I also just updated an existing IT by removing the build dep. I don't think the fix needs it's own IT.

AndyGee:
Hmm, not that simple it seems, without digging in, is there a step thad does magic on the invoker.properties?

src/it/add-third-party-no-encoding/invoker.properties
```diff
@@ -19,5 +19,5 @@
 # <http://www.gnu.org/licenses/lgpl-3.0.html>.
 # #L%
 ###
-invoker.goals=clean license:add-third-party
+invoker.goals=clean org.codehaus.mojo:license-maven-plugin:@pom.version@:add-third-party
```
ppalaga: I think you want `${project.groupId}:${project.artifactId}:${project.version}:add-third-party` see https://maven.apache.org/plugins/maven-invoker-plugin/run-mojo.html#invokerPropertiesFile

AndyGee:
That does it - no more build dep required for a command line - I'd not remove the legacy code in this request, but I'd agree it should be removed through new request/issue.

AndyGee:
The CI jobs are actually passing, but need an update to go green.

ppalaga:
I do not see the CI passing. I restarted Appveyor on Java 8 and I am investigating how to fix the broken Java installation in TravisCI.

ppalaga:
The master is now passing the whole OS x Java version matrix. Could you please rebase and also squash into a single commit?

AndyGee:
The CI is flaky

ppalaga:
Yeah, the timeouts on Appveyor do not look good.

ppalaga:
I cleaned the caches on TravisCI and restarted. Let's see if it helps.

ppalaga:
Maven Central timeouts seems to be the cause of the random failures on the CI. AppVeyor passed for this PR, TravisCI passed here: https://github.com/mojohaus/license-maven-plugin/pull/350

Issue #312, opened 2019-03-26
FileAlreadyExists exception when organizeLicensesByDependencies is true and a license is cached


srdo:
Trying to download licenses with organizeLicensesByDependencies set to true.

```
[WARNING] Unable to retrieve license from URL 'http://www.apache.org/licenses/LICENSE-2.0.txt' for dependency 'org.apache.commons:commons-lang3': /mnt/c/Users/Esran/Documents/NetbeansProjects/storm/target/generated-resources/licenses/org.apache.commons.commons-lang3_the_apache_software_license,_version_2.0
[DEBUG] 
java.nio.file.FileAlreadyExistsException: /mnt/c/Users/Esran/Documents/NetbeansProjects/storm/target/generated-resources/licenses/org.apache.commons.commons-lang3_the_apache_software_license,_version_2.0
    at sun.nio.fs.UnixCopyFile.copy (UnixCopyFile.java:551)
    at sun.nio.fs.UnixFileSystemProvider.copy (UnixFileSystemProvider.java:253)
    at java.nio.file.Files.copy (Files.java:1274)
    at org.codehaus.mojo.license.AbstractDownloadLicensesMojo.downloadLicenses (AbstractDownloadLicensesMojo.java:1082)
```

The broken code seems to be https://github.com/mojohaus/license-maven-plugin/blob/license-maven-plugin-1.19/src/main/java/org/codehaus/mojo/license/AbstractDownloadLicensesMojo.java#L1082

I'm not sure what that code is supposed to do, but if the target file is supposed to be overwritten, the copy call should set CopyOption.REPLACE_EXISTING.

ppalaga:
Thanks for the report.

Yes, I think adding CopyOption.REPLACE_EXISTING could fix it.

Looks like we have just two integration tests with `organizeLicensesByDependencies=true" which apparently do not cover this situation. Please add a test in https://github.com/mojohaus/license-maven-plugin/tree/master/src/it that reproduces your scenario and make sure it fails before and passes after applying your fix.

Vlatombe:
This happens if a dependency with the same ga appears several times in the dependency tree with different versions. Since the current [filename mapping](https://github.com/mojohaus/license-maven-plugin/blob/master/src/main/java/org/codehaus/mojo/license/AbstractDownloadLicensesMojo.java#L1130-L1133) doesn't contain the version, the second execution fails.

PR #360, opened 2019-12-16
Fixes issue #312: Don't overwrite existing license file

Vlatombe:
When using `organizeLicensesByDependencies=true`, if you have several
dependencies with the same license, and the same dependency appears
in the dependency tree with different versions/types, then the cached
file will be copied several times over to the destination, which can
already exist.


Vlatombe:
The failing test `download-licenses-configured` fail the same for me locally on master branch...
The other one `PR-33` seems to be flaky, can't reproduce the failure locally.

batmat:
Seems like this got closed unintendedly?

olamy:
@batmat really seems to be a github bug

PR #364, opened 2020-01-15
Fix #139 - "Can't find license-maven-plugin" (#1)

### LINKED ISSUE #139: "Can't find license-maven-plugin" in 1.16 if plugin isn't configured in the POM

AndyGee:
* Fix #139 - "Can't find license-maven-plugin" when run on the command line and the plugin in not defined in the POM

Remove the POM plugin definition (which is what actually validates thi fix) - This is the test.
Use the PluginDescriptor so that the POM is not used/parsed to find the artifact.
Use the project properties for the invoker.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: java
-sudo: false
+dist: trusty
 
 # Skip the Travis CI's default install command
 install: true
--- a/src/it/add-third-party-no-encoding/invoker.properties
+++ b/src/it/add-third-party-no-encoding/invoker.properties
@@ -19,5 +19,5 @@
 # <http://www.gnu.org/licenses/lgpl-3.0.html>.
 # #L%
 ###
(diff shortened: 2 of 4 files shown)


AndyGee:
The issue was closed, but still exists.

It should not be required to define the plugin in the POM just to run the plugin from the command line,

PR #374, opened 2020-04-20
Fix #351 License override does not work for third-party dependencies with multiple licenses

### LINKED ISSUE #351: License override does not work for third-party dependencies with multiple licenses
When I'm trying to use license overriding for a library that have multiple licenses only first license from the list is overridden. 
Example:
in generated THRID-PARTY.txt

` (ALS 2.0) (GPL v2) (MPL 1.1) RabbitMQ Java Client (com.rabbitmq:amqp-client:5.1.2 - http://www.rabbitmq.com)`

When I add 

`com.rabbitmq--amqp-client--5.1.2=Apache Software License - Version 2.0`

to override-THIRD_PARTY.txt file the result is

` (Apache Software License - Version 2.0) (GPL v2) (MPL 1.1) RabbitMQ Java Client (com.rabbitmq:amqp-client:5.1.2 - http://www.rabbitmq.com)`

Should license overriding replace all the licenses specified for particular dependency?

lama0206:
As discussed in #351, the current license override overrides only the first license of a multi-license artifact. The update - also as discussed in #351 - removes/overrides all licenses and logs this action on the INFO level.


lama0206:
Figured out how to run the integration tests locally :)

This is how the log messages look like (taken from the ISSUE-351 integration test's log file:

```
[INFO] Overriding license(s) for dependency [org.glassfish.jersey.containers--jersey-container-servlet--2.30.1] with [(Eclipse Public License 2.0)], overriden license(s): [(Apache License, 2.0) (BSD 2-Clause) (EDL 1.0) (EPL 2.0) (GPL2 w/ CPE) (MIT license) (Modified BSD) (Public Domain) (W3C license) (jQuery license)]
```

src/main/java/org/codehaus/mojo/license/model/LicenseMap.java
```diff
@@ -158,19 +158,31 @@ public LicenseMap toLicenseMapOrderByName()
...
                 }
             }
         }
+
+        for ( String r : removedFrom )
+        {
+            get( r ).remove( project );
+        }
+
+        return removedFrom;
```
lama0206, line 186: Removing the entries directly yields a ConcurrentModificationException. Also, I collect the removedFrom licenses for logging.

src/it/ISSUE-351/postbuild.groovy
```diff
@@ -0,0 +1,30 @@
...
+ * You should have received a copy of the GNU General Lesser Public
+ * License along with this program.  If not, see
+ * <http://www.gnu.org/licenses/lgpl-3.0.html>.
+ * #L%
+ */
+
+file = new File(basedir, 'target/generated-sources/license/THIRD-PARTY.txt');
+assert file.exists();
+content = file.text;
+assert content.contains('(Eclipse Public License 2.0) jersey-container-servlet (org.glassfish.jersey.containers:jersey-container-servlet:2.30.1 - https://projects.eclipse.org/projects/ee4j.jersey/project/jersey-container-servlet');
```
lama0206, line 26: The pom inherited a number of different licenses - the projects website states that the code is available under the Eclipse Public License 2.0. So I took that as IT.

lama0206:
Ye, lines 178-181 don't make sense. I'll take a look at it this weekend.

src/main/java/org/codehaus/mojo/license/api/DefaultThirdPartyTool.java
```diff
@@ -624,16 +624,20 @@ public void overrideLicenses( LicenseMap licenseMap, SortedMap<String, MavenProj
...
-
                     // empty license means not fill, skip it
+                    LOG.debug( " Not overriding license for dependency [{}] because no override given", id );
                     continue;
                 }
 
-                licenseMap.removeProject( project );
+                // remove project only removes first occurrence of project from license -> project[] map.
+                List<String> removedFrom = licenseMap.removeProject( project );
+                LOG.info( "Overriding license(s) for dependency [{}] with [{}], overriden license(s): [{}]",
```
nfalco79, line 634: not good print these information to log as info. This plugin is already much verbose.

lama0206: I'm not sure if I agree - yes, it could be considered verbose yet overriding a license can be a major issue which is why I agree with digulla's comment in general.

I'd rather print one summary line (for example, "Overriding license(s) for 123 dependencies. Enable debug log for details.") iff a license has been overridden instead of silently pushing the info down to LOG.debug entirely.

What do you think?

cowwoc: I agree with @lama0206

profTwinglings:
Hi do you think this will be merged?

slawekjaranowski:
merged with master
simple cleanups

src/main/java/org/codehaus/mojo/license/api/DefaultThirdPartyTool.java
slawekjaranowski: After testing with multi module projects with long override list I think that log in debug level will be enough.

PR #397, opened 2020-11-15
#396 substring ioob on missing file for multilevel projects

### LINKED ISSUE #396: Multi-level multi-module project incorrectly resolves missing file
When using a multi-module project with multiple level of child projects, the missing file is incorrectly resolved and gives an IndexOutOfBoundsException. I have an integration test and a fix ready; I will create a pull request for this issue next.

Jurrie:
Integration test and fix for #396 (fixes #396)

### DIFF
--- a/src/it/aggregate-add-third-party-multi-level-children/child1/pom.xml
+++ b/src/it/aggregate-add-third-party-multi-level-children/child1/pom.xml
@@ -1,0 +1,45 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  #%L
+  License Maven Plugin
+  %%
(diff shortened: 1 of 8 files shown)


Jurrie:
I've switched to not-using a missing file (using the override file to specify missing licenses instead).

Actually, I think I misunderstood the missing file... It's more a template file, which gets recreated sometimes even. I was confused because of the `useMissingFile` configuration directive. Should that even be possible? Shouldn't users use the override file instead, and have the missing file be a file that gets generated over and over again?

Anyway, my build works now. I close this PR as I no longer feel confident it's an actual fix...

PR #401, opened 2021-02-24
Add links to license in HTML report

### LINKED ISSUE #212: Add links to license in HTML report
The HTML report currently only shows the license name from the POMs.

If the license has a URL in them, create a link in the HTML report with the name as text of the link.

If possible, add a list of URL mappings because some links are dead.

danthe1st:
fixes #212

This pull requests adds links to licenses if license URLs are available.


slachiewicz:
This one cannot be rebased as-is: 17 conflict hunks across the five report files. `ThirdPartyReportMojo` was renamed to `ThirdPartyReport`, the reports moved to constructor injection, and the renderers were migrated to Doxia 2, so the changes need re-implementing against the current renderer rather than replaying.

@danthe1st are you still interested in this feature? Linking license names to their URLs in the HTML report is still missing, so a fresh PR against master would be welcome - and if you would rather not, say so and I will close this.

PR #406, opened 2021-04-27
Update FreeMarker version to 2.3.31

kleyton-ohare:


kleyton-ohare:
@tchemit could you review, approve and merge?

tchemit:
Thanks for the patch :)

Issue #106, opened 2017-08-25
Add lineSeparator parameter to configuration


comdiv:
Now we are building code on mix Windows/Linux dev environment, so license plugin cause invalid delta generating CRLF on one system and LF on Linux

We want to specify that we have to use only LF

Suggest to use lineSeparator in configuration with default value from System.getProperty

ragadeeshu:
I just want to chime in, I am also facing the same problem, our builds has a tendency to fail on windows because of this.

tchemit:
I guess you are talkin about license update header mojo ?
I just add a new property **lineSeparator** to specify what you need. It is use then while generating license header.

tchemit:
@comdiv, @ragadeeshu I just deployed a snapshot with this new feature.
Could you please confirm it is working fine for you ? If you need some help to use snapshot, let me know I can help.
Thanks.

comdiv:
Nice think it closes

tchemit:
@comdiv have you try the snapshot ? I will start a release 1.16, could you please test in the staging repository when it will be ready ?
If you don't know how to do this, I can help, it is quite simple. thanks

tchemit:
@comdiv here it is :) Release 1.16 is in vote (see https://groups.google.com/d/topic/mojohaus-dev/xZoobcYx-n8/discussion)

Wykiki:
Hello, I got hard-time configuring this property.
It looks like that the `lineSeparator` property for `license-maven-plugin` expect a character LF, not a placeholder like `LF` or `\n` (in the case I want to use LF as line separator).

I do not see in the documentation how to do it, and tried with that :

```xml
<license.lineSeparator>LF</license.lineSeparator>
<license.lineSeparator>\n</license.lineSeparator>
<license.lineSeparator>0x0A</license.lineSeparator>
```

Thank for your answer !

johny2000uwb:
I have the same problem as @Wykiki too. Please add some example, how should be this value used. I have tried almost everything but each value will be printed direct into text.

PR #410, opened 2021-08-23
Handle lineSeparator conversion with tokens LF / CRLF

Wykiki:
This PR is linked to my comment on the issue #106 

We are now able to provide the lineSeparator information with :

```xml
<license.lineSeparator>LF</license.lineSeparator>
<license.lineSeparator>CRLF</license.lineSeparator>
```

I tried to do the smallest change possible to avoid breaking anything, I'm not sure that it fits the codebase as I've not spent much time looking into the project.

### DIFF
--- a/src/it/update-line-separator-test-mojo/src/main/java/crlf/org/codehaus/mojo/license/CrLf.java
+++ b/src/it/update-line-separator-test-mojo/src/main/java/crlf/org/codehaus/mojo/license/CrLf.java
@@ -1,0 +1,21 @@
+/*
+ * #%L
+ * License Test :: do NOT update!
+ * %%
+ * Copyright (C) 2010 Tony
+ * %%
+ * Fake to be removed!
+ * #L%
+ */
+
+/**
+ * this class owns a header, only license will be updated.
+ *
+ * @author tchemit dev@tchemit.fr
+ * @since 1.0
+ */
+package org.codehaus.mojo.license;
+
+public class Lf
+{
+}
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,4 @@
-*        text eol=lf
-*.cmd    text eol=crlf
+*         text eol=lf
+*.cmd     text eol=crlf
+CrLf.java text eol=crlf
+Lf.java   text eol=lf
--- a/src/it/update-line-separator-test-mojo/LICENSE.txt
+++ b/src/it/update-line-separator-test-mojo/LICENSE.txt
@@ -1,0 +1,166 @@
+		   GNU LESSER GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+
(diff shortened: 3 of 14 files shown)


Djaytan (approved):

slawekjaranowski:
Should be done in `AbstractFileHeaderMojo` we can do setter method for config property `lineSeparator`, we should also update documentation on this property.

slawekjaranowski (changes requested):
Please move to `AbstractFileHeaderMojo`

Wykiki:
Updated to correct file, but not sure if setter is correctly picked up.
I wanna write test for that, should I duplicate the existing one like what's under `src/it/update-file-header-test-mojo` for this purpose ?

slawekjaranowski:
Please also update documentation for field in Java docs 

```
    @Parameter(property = "license.lineSeparator")
    private String lineSeparator;
```
It will be used for generating description like in: https://www.mojohaus.org/license-maven-plugin/update-file-header-mojo.html#lineSeparator

slawekjaranowski:
You can duplicate `src/it/update-file-header-test-mojo` but can be simplified only for checking `lineSeparator`

Wykiki:
Tried to implement tests, but it looks like line ending got replaced by system one when either `\r\n` or `\n` characters ends up as the `lineSeparator` property.

I may have more time to dig that later, but if you have any clue where this could happen, feel free to give advice !

I suspect something around following lines : 
- https://github.com/mojohaus/license-maven-plugin/blob/master/src/main/java/org/codehaus/mojo/license/header/transformer/FileHeaderTransformer.java#L67
- https://github.com/mojohaus/license-maven-plugin/blob/master/src/main/java/org/codehaus/mojo/license/header/transformer/AbstractFileHeaderTransformer.java#L239

src/it/update-line-separator-test-mojo/src/main/java/crlf/org/codehaus/mojo/license/CrLf.java
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * this class owns a header, only license will be updated.
+ *
+ * @author tchemit dev@tchemit.fr
+ * @since 1.0
+ */
+package org.codehaus.mojo.license;
+
+public class Lf
```
slawekjaranowski, line 19: Even it is test file should follow java conventions, class should the same name as file.
Please also check other test files.

slawekjaranowski:
Please always build locally ... last change has issue reported by spotless.

slachiewicz:
I rebased this onto master and pushed a commit applying the project formatting to `Eol.java`, so `spotless:check` is clean now.

The new IT still fails, though - `update-line-separator-test-mojo` asserts

```
assert assertContains(file, content, 'Copyright (C) 2010 Tony\r\n')
```

and the updated header under `src/main/java/crlf` still comes out with LF endings, so `<lineSeparator>CRLF</lineSeparator>` is not reaching the written header. That matches your own "not working yet" note on the test commit.

@Wykiki are you still able to pick this up? The remaining gap is in the mojo rather than the test.

Wykiki:
Hello, 

Not working with Java for the moment, if anyboby feels lucky picking up the PR, I'll gladly give ownership. 

Sorry

PR #424, opened 2022-03-15
Fix #423 Don't log warning for unused overrides in remote override file.

### LINKED ISSUE #423: Unwanted warnings when using shared license override file
We are using a centralised license override file for all of our Java projects which means that every build contains a large number of log lines like:
```[WARNING] dependency [org.antlr--antlr4-runtime--4.8-1] does not exist in project.```
For some of our builds this is causing several thousand lines of warnings, which is pushing the build log dangerously close to the Travis limit.

This issue is very similar to #201 and a fix has previously been attempted in #300.

### PR DESCRIPTION
**Pull request by tpage-alfresco**

### CODE REVIEW

**Comment by tpage-alfresco** [comment]
I couldn't think of a reliable way to create an integration test that used http for an override URL.

File: src/it/ISSUE-423/invoker.properties [hunk 1]
```diff
@@ -0,0 +1,2 @@
+invoker.goals=clean install license:add-third-party
```
tpage-alfresco, line 1: I don't think `install` is necessary here.  I added it while debugging an issue.

src/it/ISSUE-423/pom.xml
```diff
@@ -0,0 +1,54 @@
...
+  <version>@pom.version@</version>
+
+  <name>License Test :: ISSUE-423</name>
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <license.verbose>true</license.verbose>
+  </properties>
+
+  <dependencies>
```
tpage-alfresco, line 41: This might be testing 2.0.0, rather than the current SNAPSHOT. If so then we need to add a pluginManagement section as in other ITs.

tpage-alfresco:
This fix has gone into a release 2.0.1.alfresco-2: https://github.com/Alfresco/license-maven-plugin/tree/license-maven-plugin-2.0.1.alfresco-2

We are hoping to revert to the official releases when they return though.

slachiewicz (approved):

PR #436, opened 2022-09-13
Fix IT test add-third-party-excluded-included

VonUniGE:
Dependabot on commit 7b3eb3d bumps JUnit to 4.13.1. 

This breaks IT test add-third-party-excluded-included (this test was probably already broken) because postbuild.groovy is checking the exact version of JUnit.

This commit bumps the JUnit version in the postbuild.groovy. The IT test add-third-party-excluded-included now succeed.

The test can be run with:
````
mvn org.apache.maven.plugins:maven-invoker-plugin:3.0.0:integration-test -Dinvoker.test=add-third-party-excluded-included
````


slachiewicz:
Thx for fixing

slachiewicz (approved):

PR #439, opened 2022-09-19
Add GH verify action

slawekjaranowski:


olamy:
maybe the PR should have passed before being merged?

slawekjaranowski:
Yes, generally I agree that build should pass.
As I wrote on dev list, I don't have time to investigate deeper here.

I hope that contributors help us with it.

This PR don't introduce new problem - only show existing.

PR #441, opened 2022-10-05
Fix two IT tests aggregate-add-third-party-*

VonUniGE:
- In the IT test aggregate-add-third-party-exclude-transitive-deps, the JUnit version is no more correct.
- In the IT test aggregate-add-third-party-simple, the Junit version and the license are no more correct.

Fix them in:
- src/it/aggregate-add-third-party-exclude-transitive-deps/postbuild.groovy
- src/it/aggregate-add-third-party-simple/postbuild.groovy


olamy:
maybe you should look at https://github.com/mojohaus/license-maven-plugin/pull/419

PR #454, opened 2023-05-02
Fix assertions in aggregate-add-third-party-exclude-transitive-deps,aggregate-add-third-party-simple

slawekjaranowski:


olamy:
https://github.com/mojohaus/license-maven-plugin/pull/451

slawekjaranowski:
and #441

PR #466, opened 2023-05-03
make jdk 1.8 minimum and switch to some sugar syntax

olamy:
Signed-off-by: Olivier Lamy <olamy@apache.org>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -106,8 +106,8 @@
 
   <properties>
     <mavenVersion>3.5.4</mavenVersion>
-    <mojo.java.target>7</mojo.java.target>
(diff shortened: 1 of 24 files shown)


slachiewicz:
switch to Java 8 done in 6c9d391f24486cc11b1c53b542073b392637148e

slachiewicz:
This no longer rebases - 23 files conflict against master.

Most of the intent also looks covered already: master is on `mojo.java.target` 8 with a separate `java11-compile` execution, and `mavenVersion` is 3.9.16 where this PR sets 3.5.4.

@olamy is there anything here still worth keeping? Otherwise I would close it.

PR #468, opened 2023-05-05
use xml unit to compare xml content

olamy:
Signed-off-by: Olivier Lamy <olamy@apache.org>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -301,6 +301,18 @@
       <artifactId>junit</artifactId>
       <scope>test</scope>
     </dependency>
+
+    <dependency>
+      <groupId>org.xmlunit</groupId>
+      <artifactId>xmlunit-core</artifactId>
+      <version>2.9.1</version>
+    </dependency>
+    <dependency>
+      <groupId>org.xmlunit</groupId>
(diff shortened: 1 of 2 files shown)


slachiewicz:
Rebased onto master the pom hunk resolves cleanly, but `ISSUE-80` then fails:

```
java.lang.ClassCastException: class org.apache.xerces.jaxp.DocumentBuilderFactoryImpl cannot be cast to
class javax.xml.parsers.DocumentBuilderFactory
	at org.custommonkey.xmlunit.XMLUnit.getControlDocumentBuilderFactory(XMLUnit.java:202)
```

master now declares `xml-apis:1.4.01` (needed for JDK 8 since the odfdom upgrade), and `addTestClassPath=true` puts it on the invoker classpath, where it collides with the Groovy script loader.

Verified: `mvn -P run-its verify -Dinvoker.test=ISSUE-80` passes on master and fails on the rebased branch.

@olamy is this still something you want to land? It needs a rebase plus a way to keep xerces off the IT classpath - or switching the assertion to `xmlunit-core` (`DiffBuilder`) instead of the legacy `XMLUnit` API.

Issue #86, opened 2017-05-10
aggregate-add-third-party not threadsafe


davidkarlsen:
The aggregate-add-third-party is not threadsafe, and forces me to run builds with --builder singlethreaded which is no fun.
Could it be made threadsafe?

http://www.mojohaus.org/license-maven-plugin/aggregate-add-third-party-mojo.html

tchemit:
@davidkarlsen , I just make it so.
I just deployed a snapshot can you test it please ?
thanks.

davidkarlsen:
It's more that I got a concurrentModificationException AFAIR - not simply the flag.

davidkarlsen:
I get this:
```
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:154)                          
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:146)                          
        at org.apache.maven.lifecycle.internal.MojoExecutor.executeForkedExecutions(MojoExecutor.java:353)          
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:198)                          
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:154)                          
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:146)                          
        at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:117) 
        at org.apache.maven.lifecycle.internal.builder.multithreaded.MultiThreadedBuilder$1.call(MultiThreadedBuilder.java:200)                                                                                                          
        at org.apache.maven.lifecycle.internal.builder.multithreaded.MultiThreadedBuilder$1.call(MultiThreadedBuilder.java:196)                                                                                                          
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)                                                 
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)                                  
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)                                                 
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)                          
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)                          
        at java.lang.Thread.run(Thread.java:748)          
[...]

Issue #354, opened 2019-09-22
`ConcurrentModificationException` running `add-third-party` goal


Stephan202:
Today I observed the following build failure, after many successful executions of the `add-third-party` goal in various large (concurrent) Maven builds:
```
[ERROR] Failed to execute goal org.codehaus.mojo:license-maven-plugin:2.0.0:add-third-party (check-third-party-licenses) on project some-internal-project: could not init goal AddThirdPartyMojo for reason : null: ConcurrentModificationException -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.codehaus.mojo:license-maven-plugin:2.0.0:add-third-party (check-third-party-licenses) on project some-internal-project: could not init goal AddThirdPartyMojo for reason : null
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:215)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:156)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:148)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)
    at org.apache.maven.lifecycle.internal.builder.multithreaded.MultiThreadedBuilder$1.call (MultiThreadedBuilder.java:190)
    at org.apache.maven.lifecycle.internal.builder.multithreaded.MultiThreadedBuilder$1.call (MultiThreadedBuilder.java:186)
    at java.util.concurrent.FutureTask.run (FutureTask.java:264)
    at java.util.concurrent.Executors$RunnableAdapter.call (Executors.java:515)
    at java.util.concurrent.FutureTask.run (FutureTask.java:264)
    at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1128)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:628)
    at java.lang.Thread.run (Thread.java:834)
Caused by: org.apache.maven.plugin.MojoExecutionException: could not init goal AddThirdPartyMojo for reason : null
    at org.codehaus.mojo.license.AbstractLicenseMojo.execute (AbstractLicenseMojo.java:185)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:137)
[...]

PR #474, opened 2023-05-12
Threadsafe mojos

timw:
Make all Mojos threadsafe (some had already been marked as thread safe in previous work).

This primarily addresses #86, where the static artifact cache in `DefaultThirdPartyHelper` could be concurrently read/updated.
This patch retains the lazy initialisation of the cache, using the singleton holder approach, and I updated the similar instantiation in `SpdxLicenseList` to use the same pattern to be consistent (instead of the double checked locking).

I've reviewed the fields used by all the other mojos that are newly marked as thread-safe, and they all look OK.
Mainly this means they don't make unsafe use of statics or other potentially shared data structures (there are plenty of areas where the Mojos would not be safe for a single instance to be used on multiple threads, but the default Mojo instantiation strategy should preclude those cases)


slachiewicz (approved):

olamy:
@timw Thanks for the PR. Everything looks great. 
Another PR related to concurrent build has been merged. https://github.com/mojohaus/license-maven-plugin/issues/354
Can you please rebase and fix the conflicts?
Thanks

slawekjaranowski:
@timw I think you can go with **first** option.

I tried to search who use api package:
https://github.com/search?q=%22import+org.codehaus.mojo.license.api.ThirdPartyHelper%22+language%3AJava&type=code

and I don't see a reason to have strict `api` without breaks ...

I consider that it is rather a util than api - but for me is for internal use only. 
There was never documentation published about such api.

Issue #517, opened 2023-10-17
Support inception year only copyright statements


d-ryan-ashcraft:
Some legal experts prefer inception year to be the only year included in the copyright statement.  Currently, this option is not supported by the `maven-license-plugin`.  To help conform to this legal perspective, it would be useful if the plugin could optionally ignore the last year when generating the copyright statement's year value.

Current result:
```
Copyright (C) 2018-2023 President Business
```

Desired optional result:
```
Copyright (C) 2023 President Business
```

PR #518, opened 2023-10-17
#517: optionally support copyright values that only contain inception year rather than year range

d-ryan-ashcraft:
… year rather than year range


src/main/java/org/codehaus/mojo/license/AbstractLicenseNameMojo.java
```diff
@@ -180,6 +180,14 @@ public abstract class AbstractLicenseNameMojo extends AbstractLicenseMojo {
     @Parameter
     protected Map<String, String> extraTemplateParameters;
 
+    /**
+     * Ignore the last date for copyright year range, preferred by some legal perspectives.
```
slawekjaranowski, line 184: I'm not sure if reason should be given here - end of sentence ` ... preferred by some legal perspectives`

We can simply say what parameter does.

PR #538, opened 2023-12-11
Support loading mutliple licenseConfigFiles

tisoft:
This PR contains several changes. If you prefer, I can also make individual PRs for each.

Main features:
* allow explicit include/exclusion of Modules in the aggregate mojo
  * not all out modules are client deliverables, we include only those in the license aggrgation
* allow to load licenseConfigFile from URLs and classpath
  * we provide a company wide config file in a JAR, that can be loaded from classpath
* allow to load multiple licenseConfigFiles
  * we use this to have a company wide config file, that can be augmented with project specific files

Additionally I enabled this to build on jitpack, for an easy possibility to test unreleased versions and I fixed the maven wrapper binary.


slawekjaranowski:
Thanks for proposition.

- First I prefer to deploy snapshot version to oss.sonatype.org - I can do it

Pleas a separate PR for each feature.

slawekjaranowski:
Deploy will be added in: #539

slachiewicz:
Two of the commits here are now empty on a rebase - `license.excludedModules` and `license.includedModules` landed in 2.6.0 and are handled by `ModuleHelper.getFilteredModules`.

What is left is the `licensesConfigFile` / resource-manager work, which still conflicts in `AbstractDownloadLicensesMojo` and uses `@Component`; the codebase moved to constructor `@Inject`. The branch also carries a few commits that look fork-local (jitpack IT skip, groupId change) and should not land upstream.

@tisoft could you rebase on master with just the multiple-`licenseConfigFiles` change? Happy to review it then.

PR #568, opened 2024-05-23
Parallel processing of files

mkarg:
Projects consisting of possibly thousands of files can take long time to get processed, mostly because file operations are rather slow (compared to in-memory operations), so it makes sense to speed up things if possible without breaking backwards compatibility.

A potential solution is to use several threads, so while the first thread is still blocked by an I/O operation, another thread can start processing the next file already. With multi-core CPUs being ubiquitous these days, such a rather small change can improve the build time considerably, in particular for huge projects with thousands of files and / or slows disks (like typically on NTFS). For single-core CPUs the drawback is neglectable.

*NB: This PR mimics the solution of https://github.com/mojohaus/build-helper-maven-plugin/pull/201*

### DIFF
--- a/src/main/java/org/codehaus/mojo/license/AbstractFileHeaderMojo.java
+++ b/src/main/java/org/codehaus/mojo/license/AbstractFileHeaderMojo.java
@@ -24,16 +24,17 @@ package org.codehaus.mojo.license;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.nio.file.Files;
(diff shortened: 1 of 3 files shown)


mkarg:
Converted to draft as there are still test failures.

mkarg:
@slachiewicz @slawekjaranowski @khmarbaise Kindly requesting your review. :-)

slachiewicz (approved):

slawekjaranowski (approved):

Issue #606, opened 2024-11-14
licenseName can not be null, nor empty


vitalijr2:
I run the plugin on my project, there no any configuration for this plugin.

The POM contains inceptionYear and license info like this:
```xml
...
  <groupId>io.github.john.supergroup</groupId>
  <inceptionYear>2024</inceptionYear>
  <licenses>
    <license>
      <name>Apache 2</name>
      <url>https://www.apache.org/licenses/LICENSE-2.0.txt</url>
    </license>
  </licenses>
  <modelVersion>4.0.0</modelVersion>
...
```

The goal `check-file-header` [has three required parameters](https://www.mojohaus.org/license-maven-plugin/check-file-header-mojo.html): inceptionYear, organizationName and projectName.

I use next command line:
```bash
./mvnw license:check-file-header -Dlicense.inceptionYear=2024 -Dlicense.organizationName=John\ Doe -Dlicense.projectName=some-project.
```

It fails with message:
```log
[ERROR] Failed to execute goal org.codehaus.mojo:license-maven-plugin:2.4.0:check-file-header (default-cli) on project some-project: could not init goal CheckFileHeaderMojo for reason : licenseName can not be null, nor empty -> [Help 1]
```

Looks like `licenseName` is a required parameter too and should be set manually in the configuration or CLI parameters.

Because when I run
```bash
./mvnw license:check-file-header -Dlicense.licenseName=apache_v2 -Dlicense.inceptionYear=2024 -Dlicense.organizationName=John\ Doe -Dlicense.projectName=some-project
```
it finish successfully.

PR #607, opened 2024-11-14
licenseName is required

vitalijr2:
Resolves issue #606: `licenseName` is required

### DIFF
--- a/src/main/java/org/codehaus/mojo/license/AbstractLicenseNameMojo.java
+++ b/src/main/java/org/codehaus/mojo/license/AbstractLicenseNameMojo.java
@@ -227,9 +227,6 @@ public abstract class AbstractLicenseNameMojo extends AbstractLicenseMojo {
     // ----------------------------------------------------------------------
 
     protected License getLicense(String licenseName, boolean checkIfExists) {
-        if (StringUtils.isEmpty(licenseName)) {
-            throw new IllegalArgumentException("licenseName can not be null, nor empty");
-        }
         if (licenseStore == null) {
             throw new IllegalStateException("No license store initialized!");
         }
--- a/src/main/java/org/codehaus/mojo/license/AbstractLicenseNameMojo.java
+++ b/src/main/java/org/codehaus/mojo/license/AbstractLicenseNameMojo.java
@@ -114,7 +114,7 @@ public abstract class AbstractLicenseNameMojo extends AbstractLicenseMojo {
      * @parameter property="license.licenseName"
      * @since 1.0
      */
-    @Parameter(property = "license.licenseName")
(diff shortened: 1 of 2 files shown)


src/main/java/org/codehaus/mojo/license/AbstractLicenseNameMojo.java
```diff
@@ -114,7 +114,7 @@ public abstract class AbstractLicenseNameMojo extends AbstractLicenseMojo {
      * @parameter property="license.licenseName"
      * @since 1.0
      */
-    @Parameter(property = "license.licenseName")
+    @Parameter(property = "license.licenseName", required = true)
```
slawekjaranowski, line 117: We can also simplify method - `getLicense`

vitalijr2: Thanks, good point.

* Remove empty check from getLicense(...) as we have same check in the license repository.
* Replace `isEmpty` with `isBlank` because a user can set blank name through a command line parameter. It fails anyway but with `isBlank` the error message better describes what the issue is.

src/main/java/org/codehaus/mojo/license/model/LicenseRepository.java
```diff
@@ -165,8 +165,8 @@ public License[] getLicenses() {
 
     public License getLicense(String licenseName) {
         checkInit("getLicense");
-        if (StringUtils.isEmpty(licenseName)) {
-            throw new IllegalArgumentException("licenceName can not be null, nor empty");
+        if (StringUtils.isBlank(licenseName)) {
+            throw new IllegalArgumentException("licenceName can not be null, nor blank");
```
slawekjaranowski, line 169: This also should not happen - Maven take care about required parameters.

src/main/java/org/codehaus/mojo/license/AbstractLicenseNameMojo.java
slawekjaranowski: As I know required parameter cannot be null nor blank

src/main/java/org/codehaus/mojo/license/model/LicenseRepository.java
vitalijr2: Нes, a user cannot set an empty license name in the configuration.

But this is possible if the license name is passed via a command line parameter. And I left this check for just this case.
I don't know why the user would do such strange manipulations. Maybe this is over-engineering?

vitalijr2: Remove this check

slawekjaranowski (approved):

PR #620, opened 2025-01-17
Implement Optional Parameters to include and exclude submodules

MrBanana05:


slawekjaranowski:
A unit tests and at least one IT - will be appreciated

MrBanana05:
Hi, thanks for the feedback! I finally found some time to write the tests. Please let me know if you need anything else.

src/main/java/org/codehaus/mojo/license/AggregateDownloadLicensesMojo.java
```diff
@@ -100,6 +101,22 @@ public class AggregateDownloadLicensesMojo extends AbstractDownloadLicensesMojo
     @Parameter(property = "reactorProjects", readonly = true, required = true)
     private List<MavenProject> reactorProjects;
 
+    /**
+     * An array of modules to include in analysis.
+     *
+     * @since 2.6.0
+     */
+    @Parameter(property = "license.includedModules")
+    private String[] includedModules;
```
slawekjaranowski, line 110: can be `List<String>`  but it is no issue

slawekjaranowski (approved):

PR #622, opened 2025-02-09
Added advanced data formatting and sorting

Master-Code-Programmer:
Hi, I made a bigger update because I was missing some clarity and readability in the license reports. With big projects and an enormous amount of (transitive) dependencies I got lost, so I added these settings.
Feel free to comment on whatever changes you need, to accept this pull-request.

The essential changes are:
* I added a `DataFormatting` option to `AbstractDownloadLicensesMojo` to format the data. The specific settings are:
   * `orderBy`, to order / sort the output by 4 possible values:
     * Dependency Name
     * Plugin ID
     * License Match (OK license, Problematic license, Forbidden license, Unknown license)
     * License Name
    I marked the option `sortByGroupIdAndArtifactId` in `AbstractDownloadLicensesMojo` therefore as deprecated.
   * `problematicLicenses`, to list problematic licenses, which get highlighted in a special color in the XML or ODS report. As also a different sort order when using `orderBy` License-Match
   * `okLicenses`, to list licenses which are explicitly allowed. They get highlighted in a special color in the XML or ODS report. As also a different sort order when using `orderBy` License-Match
   * Forbidden licenses just reuse the already existing `excludedLicenses`.
   * `matchedLicensesHaveBorder`. If licenses found in `excludedLicenses`, `problematicLicenses`, `okLicenses` or where `highlightUnknownLicenses` is used, have a visible border. The border makes license types better visible at a glance, but when quickly scrolling through a long list, it makes licenses look as if they were the only ones belonging to a dependency if it has multiple licenses.
   * `highlightUnknownLicenses`. If all unknown licenses should be highlighted. Unknown means: There is no entry for a license in `excludedLicenses`, `problematicLicenses`, `okLicenses`.
   * `skipDevelopers` just skip the mentioning of developers in the license reports. They are most likely irrelevant for the license report, but are the main reason one dependency entry can get far to many lines. Which makes the report just much harder to read and grasp.
   * Tests for all 4 sortings are in `aggregate-download-licenses-sort-by-dependencyName`, `aggregate-download-licenses-sort-by-licenseMatch`, `aggregate-download-licenses-sort-by-licenseName` and `aggregate-download-licenses-sort-by-pluginId`.
   Example `pom.xml`:
   ```xml
   <dataFormatting>
        <orderBy>licenseMatch</orderBy>
        <problematicLicenses>
           <problematicLicense>EPL 2.0</problematicLicense>
           <problematicLicense>MIT License</problematicLicense>
         </problematicLicenses>
         <okLicenses>
           <okLicense>Apache License, Version 2.0</okLicense>
         </okLicenses>
         <matchedLicensesHaveBorder>true</matchedLicensesHaveBorder>
         <highlightUnknownLicenses>true</highlightUnknownLicenses>
         <skipDevelopers>true</skipDevelopers>
   </dataFormatting>
   ```
* `LicenseDownloader` now sends a User-Agent, so license downloads from `gnu.org` can become possible. Otherwise it sends an irritating and misleading error: "429 Too many requests". And "gnu.org" is an extremely often used license hoster.
* Bug-Fix: Not all dependencies must be JARs. This should be rarely the case, but it happens. So that is taken into account when extracting extended dependency info, and Non-JARs are skipped and therefore don't throw an exception.
* I added the blank-space-filtering I use for license name merges also to `AbstractAddThirdPartyMojo`s `includedLicenses` and `excludedLicenses`, by adding it to `StringToList`. So you can not only write (the old way):
  `<includedLicenses>licenseA|licenseB</includedLicenses>`, but also:
  ```xml
  <includedLicenses>licenseA
    |licenseB
    |licenseC
  </includedLicenses>
  ```
  I did this not only for readability, but also since some aggressive XML-Formatters enforce the latter format if the lines get too long anyway, and expect the parsers to handle this.
* I added, as JavaDoc, to `AbstractDownloadLicensesMojo`s `licenseMerges` all mainly used license-name variations. This is always a huge waste of time, if you set up a new project from scratch and need to find out again all standard name-variations of licenses. I don't want to add that as a default, since I consider that to be maybe too much "magic", and it should be very transparent and intentionally decided which license names mean really the same. So this works as an _extremely_ convenient time-saving suggestion.


Master-Code-Programmer:
@slawekjaranowski I would've donated, like a "beer", to you, but w/o a credit-card I found no way. You could add a Patreon account.

src/it/aggregate-download-licenses-shared/convert.sh
slawekjaranowski, line 1: only used in readme ....

src/it/aggregate-download-licenses-shared/shared.groovy
```diff
@@ -0,0 +1,235 @@
...
+            ", groupId='" + groupId + '\'' +
+            ", artifactId='" + artifactId + '\'' +
+            ", version='" + version + '\'' +
+            ", licenses='" + licenses + '\'' +
+            '}'
+    }
+}
+
+// -------------- Sort by name -------------------
+static void checkResultingLicensesXml(Logger log, File basedir, String expected)
```
slawekjaranowski, line 100: why so such complicated method .... 
why not use groovy `xmlslurper` - https://groovy-lang.org/processing-xml.html

slawekjaranowski (changes requested):

src/it/aggregate-download-licenses-shared/convert.sh
Master-Code-Programmer: First: Sorry for the delay, I hope I got now much more time to get this PR ready.

The script is there to make the XML files easily readable, otherwise it's not obvious how the XSLT could be used to get a clear view of the sorting in the template XML files.

I needed some time to find it out, and found it also hard to read the _raw_ XML files, to check if the generated XML files were actually correct. I don't want to test against incorrect data.
Here is an example for the Plugin-ID sorting:
grafik

src/it/aggregate-download-licenses-shared/shared.groovy
Master-Code-Programmer: Thanks, that is an interesting solution, the code is now much more compact. I pushed that.
Although: I researched the topic and found out that for my task I should use the "XML Parser", since I need to read always the whole XML document anyway.

Master-Code-Programmer: @slawekjaranowski Hi, I just wanted you to remember / inform, that you could review my changes.

slawekjaranowski: it is still on my list ... but I need to understand it

Master-Code-Programmer:
These "new" commits are just a rebase, to keep the PR up to date.

src/it/aggregate-download-licenses-sort-by-dependencyName/pom.xml
```diff
@@ -0,0 +1,211 @@
...
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+      <version>2.15.2</version>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-core</artifactId>
+      <version>2.15.2</version>
+    </dependency>
```
slawekjaranowski, line 53: Maybe dependencies list for testing can be shorter ...?

slachiewicz:
This conflicts with master. The main obstacle is `CalcFileWriter`, which moved to `src/main/java11` in #684 when odfdom went to 0.12.0 and ODS generation started requiring JDK 11+, so the formatting changes need re-applying at the new location.

@Master-Code-Programmer is this still active? If you can rebase it onto current master we can take another look, otherwise I will close it as stale.

Master-Code-Programmer:
Sorry, I didn't pay much attention to this PR, also when I did my rebase, I first thought you deleted the Calc export, instead of moving it.

I will look into this.

Master-Code-Programmer:
I rebased, but I had many merge conflicts, and need to resolve them in the next days, the current state isn't working.
Also: Apache POI 6 is on the doorsteps (for Excel export), and will be released soon. It also requires Java 11.

But if you just need to have a JDK 11+ on your project, and Maven loads "automagically" the correct implementation, I'm fine w/ that. That sounds _very_ convenient.

slachiewicz:
Thanks for this, and sorry it sat so long. Rather than ask you to carry an 18-month branch through another rebase, I have rebuilt the feature against current master and split it across five PRs: #730 ordering, #731 the license lists, #732 the Excel and Calc highlighting, #733 `skipDevelopers`, and #734 the whitespace handling in `includedLicenses`. Every commit credits you as co-author, and the earlier #728 and #729 came out of this PR too. This one will close as superseded once they land, and I would rather you reviewed them first.

Two things changed from your design.

The options are flat mojo parameters rather than a nested `<dataFormatting>` element. That way `-Dlicense.orderBy=licenseMatch` works from the command line, and no new public config class enters the plugin API. `dependencyPluginId` became `dependencyGav`, which is what it orders by.

The forbidden list is its own parameter instead of reusing `excludedLicenses` from `add-third-party`. That parameter also fails the build, and failing on a forbidden license is a separate feature from colouring one. It is worth having; please open it on its own if you want it.

Three defects from this branch that were worth catching, in case they are useful to you:

* `execute()` called `sortProjectLicenseInfo(...)` unconditionally, which dereferences `dataFormatting`. A nested complex-object parameter is null when the POM omits the element, so that was an NPE for everyone who did not configure `<dataFormatting>`. All four new ITs configure it, which is why they passed.
* The highlighting was gated on `getOrderBy() != none`, so an explicit `<orderBy>none</orderBy>` turned the colours off while an unset `orderBy` turned them on. Highlighting and ordering now have independent switches.
* `@Parameter` on the fields of a nested config class has no effect; `property` and `defaultValue` are honoured only on Mojo fields.

One correction to your description, which changed how #729 was written: non-archive dependencies never threw. `ZipException` was already caught, so a `.pom` or native dependency only logged a stack trace at WARN. #729 catches it separately and keeps the artifact info, rather than skipping by a `jar`/`zip` extension whitelist, which would also skip `war`, `aar` and `hpi`.

On the ITs, closing out @slawekjaranowski's earlier points: none of the five PRs adds an IT project. The ordering is proven by making the expected order the reverse of the resolution order in ITs that already existed, and the highlighting by writing a spreadsheet and reading the colours back in a unit test.

*This comment was created with AI assistance.*

Master-Code-Programmer:
Big thanks for all your work. No worry that it sat so long. That was my fault, not rewriting the code, the requests of [slawekjaranowski](https://github.com/slawekjaranowski) were reasonable.

I already finished the rebase, and now that I have the possibility to use the latest ODF Toolkit, I started looking into the possibility to replicate all features of the Excel export into the Calc export (there is an open TODO for grouping, for example). But I will stop that, and make a new PR for that (if I get it running at all).

First, I welcome your corrections, from someone which has much more project and Maven knowledge than I do. I welcome each PR and will review it.

Thanks again.

PR #650, opened 2025-07-11
Check project license goal

sewerwal:
Introduces new goal check-project-license to check existing of license file and optionally fail build.


sewerwal:
> Why not use https://maven.apache.org/enforcer/enforcer-rules/requireFilesExist.html ?

sure i can use it, but this plugin generates license file so can also check if it exists, similar to goal license:check-file-header

src/main/java/org/codehaus/mojo/license/CheckProjectLicenseMojo.java
```diff
@@ -0,0 +1,100 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * The goal to check if there are license files in project.
+ *
+ * @author swalendzik
+ * @since 2.7
+ */
+@Mojo(name = "check-project-license", threadSafe = true)
+public class CheckProjectLicenseMojo extends AbstractLicenseNameMojo {
```
slawekjaranowski, line 43: `AbstractLicenseMojo`  should be enough here

src/main/java/org/codehaus/mojo/license/CheckProjectLicenseMojo.java
```diff
@@ -0,0 +1,100 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(CheckProjectLicenseMojo.class);
+
+    // ----------------------------------------------------------------------
+    // Mojo Parameters
+    // ----------------------------------------------------------------------
+
+    /**
+     * Project license file to check.
+     *
+     * @required
```
slawekjaranowski, line 53: there is no tag `@required`

src/main/java/org/codehaus/mojo/license/CheckProjectLicenseMojo.java
```diff
@@ -0,0 +1,100 @@
...
+     * @since 2.7
+     */
+    @Parameter(property = "license.licenceFile", defaultValue = "${basedir}/LICENSE.txt")
```
slawekjaranowski, line 56: `File` parameters are always aligned by current project, so `${basedir}` is not needed.

src/main/java/org/codehaus/mojo/license/CheckProjectLicenseMojo.java
```diff
@@ -0,0 +1,100 @@
...
+    protected File licenseFile;
```
slawekjaranowski, line 57: why `protected`?

src/main/java/org/codehaus/mojo/license/CheckProjectLicenseMojo.java
```diff
@@ -0,0 +1,100 @@
...
+
+    @Override
+    protected void doAction() throws Exception {
+        if (!licenseFile.exists()) {
+            String message = "There is module without license " + this.licenseFile.getAbsolutePath()
+                    + ". Use license:update-project-license to generate";
+            if (failOnMissingLicense) {
+                throw new MojoFailureException(message);
+            } else {
+                LOG.warn(message);
```
slawekjaranowski, line 96: simple `getLog().warn(...)` cn be used here

slawekjaranowski (changes requested):
Please add at least one IT, see at `src/it` for examples.

src/main/java/org/codehaus/mojo/license/CheckProjectLicenseMojo.java
sewerwal: changed

src/main/java/org/codehaus/mojo/license/CheckProjectLicenseMojo.java
sewerwal: removed

src/main/java/org/codehaus/mojo/license/CheckProjectLicenseMojo.java
sewerwal: changed to private

slawekjaranowski (approved):

PR #718, opened 2026-07-21
Feat: Allow easier Reusing / Normalization of License Texts

MrBanana05:
We've recently reviewed our process for downloading License Texts and it's result. We noticed that there we're tons of unreadable license texts, caused by maintainers of Third Party Libraries referencing websites with tons of content in addition to the License Texts themselves. Most of these libraries are also using some common license. We reviewed the currently existing options for overriding URLs based on License Names and only found <licensesConfigFile />, which did not satisfy our needs, since <matchLicenses /> matches the order of the licenses and is intended for Dependencies, not License Names. This is why we decided to implement licenseNameUrlOverrides, to override URLs based on the name of the license. This allows us to define regex to match multiple variations of the same license name and override the URL used for downloading the License Text.


slachiewicz (commented):
Thanks for this - the use case (pointing license downloads at an internal mirror) is a real one. A few things to work through before it can land.

### The override never reaches `ProjectLicense`, and the cache is keyed on the POM URL

`downloadLicenses` uses the overridden URL as a local variable, and `handleResult` only writes `<file>` back. `licenses.xml` therefore still records the POM URL, and on the next run `execute()` primes the URL cache from it (`AbstractDownloadLicensesMojo.java:928`).

Until now every rewrite was keyed on the URL itself, so two dependencies declaring the same `<url>` always resolved to the same download. Keying on the license *name* breaks that:

- dep A declares `Apache License, Version 2.0` + `http://www.apache.org/licenses/LICENSE-2.0.txt`, matches an override pointing at an internal mirror
- dep B declares `The Apache Software License, Version 2.0` with the same POM URL, and does not match

First build is correct. Second build (the default `forceDownload=false`) primes the cache with A's POM URL against the mirror's file, dep B hits that entry, and B ends up recorded with the mirror's license text. A wrong license attributed to a dependency, appearing only from the second build onward.

### The overrides file is not validated

Four ways a small mistake goes unreported:

1. A matching entry whose `<url>` is missing or empty returns null, and `matchingUrlsOnly && isBlank(nameOverriddenUrl)` then reports *"No URL for license at index 0 in dependency ..."* - so a dependency that has a perfectly good POM URL is reported as having none, and fails the build under `errorRemedy=failFast`. A blank override should fall back to `currentUrl`.
2. An entry with a mistyped element name (`<licenceNamePattern>`) parses with `pattern == null` and is skipped by the `getPattern() != null` guard, with nothing in the log.
3. `getTextContent()` is not trimmed, so a pretty-printed file (value on its own line) yields a pattern containing newlines and indentation - `Pattern.matches` then never fires, silently. The same on `<url>` makes `new URI(url)` throw and reports "invalid license URL".
4. A `licenseNameUrlOverridesFile` that does not exist is treated as an empty list. For an air-gapped build, that is the difference between "overrides applied" and "silently downloaded everything from the internet".

### XXE

`LicenseNameUrlOverridesReader` builds a `DocumentBuilderFactory` without disabling DOCTYPE/external entities before parsing a user-supplied file. A file carrying `<!ENTITY x SYSTEM "file:///...">` is expanded during the build. Please set `disallow-doctype-decl` (or `XMLConstants.FEATURE_SECURE_PROCESSING` plus the external-entity features).

### The documented example cannot match

Matching uses `Matcher.matches()`, so the pattern must match the whole name, but the example in the javadoc is

```xml
<licenseNamePattern>(?i)apache.*2\.0</licenseNamePattern>
```

which does not match `The Apache Software License, Version 2.0` - the name most ASF artifacts on Central actually declare. Either the example needs a leading `.*`, or matching should use `find()`.

### The IT passes when the feature is broken

`src/it/download-licenses-name-override/postbuild.groovy` asserts on `Eclipse Public License - v 1.0`, which is also the heading of the genuine `epl-v10.html`. If the override stops applying, the plugin downloads the real page and the assertion still succeeds. Assert on the discriminating marker instead - the placeholder text in `src/license/epl-1.0.txt`.

### Error messages blame the wrong file

The catch blocks around the download still say *"POM for dependency X has an invalid license URL"*, but the URL can now come from the overrides file. A typo there fails the build naming a dependency POM the user cannot change, and never names `licenseNameUrlOverridesFile`.

### Do we need the new file at all?

Two questions on the shape rather than the code:

- `licensesConfigFile` already does name-regex to URL replacement. `DependencyMatcher` compiles `groupId`/`artifactId`/`version` as case-insensitive regexes and `matchLicenses` matches the license list, and `matchers.replaceMatches(...)` runs before downloading (`AbstractDownloadLicensesMojo.java:914`). An entry with `.*` for groupId/artifactId and your name pattern under `<matchLicenses>` looks like it would express the same override today. If it does not cover your case, that gap is worth stating in the PR - it may be a smaller change to `LicenseMatchers`.
- If a new parameter is the right answer, a `@Parameter List<LicenseNameUrlOverride>` populated by Maven would remove the 95-line DOM reader and most of the failure modes above at once, and would match the neighbouring `licenseUrlReplacements`, `licenseUrlFileNames` and `licenseUrlFileNameSanitizers`, which are all plain POM-configured collections. Users would get schema validation, IDE completion and `${...}` interpolation for free.

MrBanana05:
Thanks for your feedback! Before addressing your code review, I'd like to discuss your questions about whether this feature is actually needed.

- `licensesConfigFile` does perform a name-regex URL replacement. Currently, however, it also enforces the order of the licenses in the pom.xml of the dependency. Therefore, if I were to add an override for licenses a and b, and dependency x lists the licenses as a and b, the licenses would be overridden. If license y lists them as license b and a, the licenses would not be overridden. This is an easy fix. However, if dependency z only lists license a or b, the license would not be overridden. This would result in many very similar overrides for a few licenses. In our opinion, the `<licensesConfigFile/>` is designed for overrides based on `groupId`/`artifactId`/`version` rather than license names. License names seem to be an optional, additional filter for dependencies. It does not seem like the `<licensesConfigFile/>` is not intended for this kind of configuration / use case. Overriding license URLs based solely on license names simplifies our configuration significantly. Otherwise, the overrides would be unmaintainable. For context: we have around 240 dependencies in one of our delivery units alone.
- We decided to use a dedicated file because it allows us to reuse the same file across multiple modules and parents. However, having multiple parents may be a unique to us, and we're actually also working on improving this. I'm open to removing the dedicated file and using `@Parameter List<LicenseNameUrlOverride>` instead.

Please let me know if you're open to adding the `LicenseNameUrlOverride` as a new feature and I'll rework the current implementation to use `@Parameter List<LicenseNameUrlOverride>` instead.

slachiewicz:
Yes - go ahead with `@Parameter List<LicenseNameUrlOverride>` and drop the separate file.

Your analysis of `licensesConfigFile` is right, and there is a second reason it is the wrong tool: `LicenseMatchers.replaceMatches` also sets `approved=true` and clears the dependency's downloader messages, so using it as a URL override silently suppresses error reporting for those dependencies. `LicenseListMatcher.matches` on top of that requires the list sizes to be equal, so a match is over the whole license list, positionally - never over a single name. The gap is real.

On reuse across parents: put the `<licenseNameUrlOverrides>` block in `pluginManagement` in your shared parent. That gets you what the file was giving you, plus `${...}` interpolation and IDE completion.

The rework moots most of my review - dropping `LicenseNameUrlOverridesReader` removes the XXE finding and validation points 2 to 4 (mistyped element names, untrimmed text content, missing file). Two things survive it.

**The URL cache.** This is the functional blocker, and the parameter change does not fix it. `handleResult` writes only `<file>` back, so `licenses.xml` keeps the POM URL while the file came from the override URL. On the next build with the default `forceDownload=false`, `execute()` primes the cache from that pair (`AbstractDownloadLicensesMojo.java:872`), and any other dependency declaring the same POM URL whose name does *not* match an override hits that entry and gets recorded with the overridden text. A wrong license text attributed to a dependency, appearing only from the second build onward. Smallest fix: apply the overrides when priming as well - `overrideIfNecessary(lic.getName(), lic.getUrl())` in that loop - so the primed keys match the lookup keys. `LicenseSummaryReader` does populate `<name>`, so it is available there.

**Whole-string matching.** Keep `Matcher.matches()`; that is what `LicenseMatchers` uses throughout, with `CASE_INSENSITIVE`. Fix the javadoc example instead - as written it cannot match `The Apache Software License, Version 2.0`.

The rest, briefly:

- With a `@Parameter` you can validate at the start of `execute()` and fail fast on a blank `<url>` or an uncompilable pattern, rather than skipping the entry.
- Assert on the placeholder text from `src/license/epl-1.0.txt` in `postbuild.groovy`. The current assertion on `Eclipse Public License - v 1.0` also passes against the genuine page, so the IT stays green if the override stops applying.
- The "invalid license URL" messages should name the override that produced the URL, not the dependency POM the user cannot change.
- Log applied overrides at INFO. Name-keying is easier to get wrong than URL-keying, and an over-broad regex is otherwise invisible.
- `@since` should match what actually ships; master is `2.7.2-SNAPSHOT`.

*This comment was created with AI assistance.*

PR #729, opened 2026-08-23
Stop warning about dependencies that are not archives

slachiewicz:
A dependency whose file is not a ZIP archive — `pom` type, or a native `.exe`, `.so`, `.dylib` — reaches `ZipFile` like any other, and the resulting `ZipException` was logged at WARN with a full stack trace once per dependency. Nothing is wrong there, so it is now a debug message; the artifact keeps its `ExtendedInfo` either way.

Behaviour is otherwise unchanged: the exception was already caught, so this is log noise rather than a failure. `LicensedArtifactTest` is new and pins both paths, including the JAR one that has to keep reading the manifest and `META-INF/LICENSE.txt`.

Extracted from #622, where @Master-Code-Programmer reported the case. That PR skipped non-archives by an extension whitelist of `jar` and `zip`, which would also skip `war`, `aar` and the other archive packagings, and returned `null` instead of the artifact info.

*This change was created with AI assistance.*

### DIFF
--- a/src/main/java/org/codehaus/mojo/license/download/LicensedArtifact.java
+++ b/src/main/java/org/codehaus/mojo/license/download/LicensedArtifact.java
@@ -34,6 +34,7 @@ import java.util.Set;
 import java.util.jar.Attributes;
 import java.util.jar.Manifest;
 import java.util.zip.ZipEntry;
+import java.util.zip.ZipException;
 import java.util.zip.ZipFile;
 
 import org.apache.commons.collections.CollectionUtils;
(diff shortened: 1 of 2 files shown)


Master-Code-Programmer:
I get your point, while I could add the missing extensions, whitelisting holds the problem of an endless, never complete list of "correct" file extensions.
But since it's bad practice to work with expected exceptions (and expensive too), instead of true / false, I suggest I add code which checks if the file has a valid ZIP header.

Would that be OK?

Master-Code-Programmer:
My PR is here: #735.

PR #740, opened 2026-08-30
Fixed skipDevelopers in CalcFileWriter

Master-Code-Programmer:
Because the ODF / ODS exporter (or LibreOffice?) copies the styles of empty cells from the row above, empty cells must be explicitly written as that: Cells w/o content or style. Otherwise you may end up with a dozen greenly backgrounded header lines.

I reintroduced that from my code into your code for the skipDevelopers part. There are still same strange behaviors with empty cells, but this solution solves so far only the newly introduced bug, which appears with skip-developers.

Calc export w/o my fix:
Calc Export without Fix
Calc export w/ my fix:
Calc Export with Fix

P.S.: I also fixed one wrong variable name and applied spotless (meaning, for a comparison "Hide whitespaces" should be selected in GitHub).

P.P.S.: I also tried to fix other already existing ODS bugs, but the ODF-Toolkit for ODS seems bad, compare to Apache POI, and far out of the scope of a small correction.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -132,7 +132,7 @@
 
     <version.httpcomponents.httpclient>4.5.14</version.httpcomponents.httpclient>
     <version.httpcomponents.httpcore>4.4.16</version.httpcomponents.httpcore>
-    <version.odfdom-java>0.12.0</version.odfdom-java>
+    <version.odfdom-java>0.13.0</version.odfdom-java>
 
     <!-- license configuration -->
     <license.useMissingFile>true</license.useMissingFile>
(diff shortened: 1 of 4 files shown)


slachiewicz:
what we should do with Java 8 build failures @Master-Code-Programmer ?

Master-Code-Programmer:
> what we should do with Java 8 build failures @Master-Code-Programmer ?

Fixed the Java 8 / 11 testing problem.

I also found a possible fix, for the other cell format duplicates you see in the right side of the screenshots. It's pretty simple, but I didn't push it because I wanted to be sure that I don't create new bugs in edge cases. I've to test that further.

_AI was used as support for this fix._

PR #745, opened 2026-09-13
Highlight licenses

Master-Code-Programmer:
Your original PR, extended with the following:

# Short form
The **LibreOffice Calc** export produces now a functional identical export to **Microsoft Office Excel**.

# Long form
* I fixed that some empty cells always copied the background from the cells above, which looked especially broken with the 1st few rows, when they copied the turqoise headers background color and thick black border.
* The cell width and height are now limited. The height is always the standard height, and while some cells width expands, they're limited and don't try to show the whole content, thereby _not_ spanning over multiple screen pages, and make the whole document hard to read.
* The 3 header rows and the 1 dependency name column are now freezed. So no matter where you scroll, you always know what each cell value stands for. _Finally_, I could get it working, with AI, but still, the AI used my ground work to make it work, and it had no fundamentally better solution. The ODF SDK is very basic.
* I replaced a deprecated call to the ODF SDK.
* Also with the AI I could add groups, which I consider very important. Before grouping I always heard the complaint that there was too much data, and it was all too confusing, with the grouping exists a very easy way to filter out all uninteresting data. For each individual use-case, with a few mouse-clicks. Just click on a minus and collapse the unwanted group.

# Non-Code fixes
* Spotless formats now also the Java 11 directories.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -404,6 +404,21 @@
         </executions>
       </plugin>
 
+      <!-- Required for Spotless to detect the Java 11 source directories. -->
+      <plugin>
+        <groupId>com.diffplug.spotless</groupId>
+        <artifactId>spotless-maven-plugin</artifactId>
+        <configuration>
+          <java combine.children="append">
+            <includes>
+              <include>src/main/java/**/*.java</include>
+              <include>src/test/java/**/*.java</include>
+              <include>src/main/java11/**/*.java</include>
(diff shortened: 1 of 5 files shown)


Master-Code-Programmer:
Sorry, for the confusing log. Tried rebasing it on master, but saw that it made it more confusing, so switched back to rebasing it on your branch.

